PR #3, opened 2021-03-17
updating docs for install/uninstall

whitleykeith:
self-explanatory

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,8 +23,7 @@ To install Airflow you simply need to fork the repo and run the following on a b
 ```bash
 # all commands are run at the root of your git repo
 # install the airflow stack and have it point to your fork of the dag code.
-# USER should be your github username, and BRANCH should be the branch of that repo. 
-./scripts/install.sh -u $USER -b $BRANCH
+./scripts/install.sh
 ```
 
 ### Getting Cluster Configuration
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -17,4 +18,19 @@ Look at [variables](./dags/openshift_nightlies/docs/variables.md) to see how var
...
 
-Refer to the [Leviathan Repo](https://github.com/whitleykeith/leviathan)
\ No newline at end of file
+To install Airflow you simply need to fork the repo and run the following on a box that has access to an openshift cluster:
+
+```bash
+# all commands are run at the root of your git repo
+# install the airflow stack and have it point to your fork of the dag code.
+# USER should be your github username, and BRANCH should be the branch of that repo. 
+./scripts/install.sh -u $USER -b $BRANCH
```
chaitanyaenr, line 27: Maybe use the code local to the fork, this way we can skip on passing the USER and BRANCH.

chaitanyaenr (commented):
Minor nit. We can track the change in a different PR as well.

README.md
mohit-sheth: ^^ as default and params if we want to use a different user/branch

whitleykeith: I'm fine with the idea but I haven't found a great way of doing it with the git config commands I've been messing around with. I've only been able to find a user's email or their human readable name like "Keith Whitley"

chaitanyaenr (approved):
LGTM

PR #7, opened 2021-04-14
Add upgrade test to the dags

chaitanyaenr:
### Description
This commit:
- Modifies the manifest for 4.7 dags to use stable releases vs nightlies.
  This is needed as the release nightlies do not have a supported upgrade
  path to release+1.
- Adds upgrade test which includes scaling up the cluster to 50 nodes,
  running cluster density with 1000 job-iterations and running the upgrade.


whitleykeith (approved):
lgtm

PR #9, opened 2021-04-28
Indexing updates, uperf workloads, and authentication fixes

whitleykeith:
### Description

* Changes indexing logic of DAGs by creating individual index tasks for every benchmark and install task. 
* Adds all uperf workloads within a parallel task group by default
* Add retries to benchmarks
* Migrate from fluentd bitnami to more performant openshift-logging fluentd
* Enforce user created password with the install script to prevent airflow being started with default credentials.


dags/openshift_nightlies/scripts/run_benchmark.sh
```diff
@@ -14,7 +14,7 @@ done
 setup(){
     mkdir /home/airflow/workspace
     cd /home/airflow/workspace
-    git clone https://github.com/cloud-bulldozer/e2e-benchmarking
+    git clone -b fix-run_comapre-dot-sh https://github.com/kedark3/e2e-benchmarking
```
amitsagtani97, line 17: We have these changes merged..can switch back to the cloud-bulldozer/e2e-benchmarking

dags/openshift_nightlies/scripts/run_benchmark.sh
```diff
@@ -33,5 +33,9 @@ setup
 cd /home/airflow/workspace
 ls
 cd e2e-benchmarking/workloads/$workload
+
+if [[ $workload == "network-perf" ]]; then
```
amitsagtani97, line 37: It will be better to run the python library install command in the e2e-benchmarking repo itself.

amitsagtani97 (changes requested):

dags/openshift_nightlies/scripts/run_benchmark.sh
whitleykeith: yep forgot to move that back, good catch

dags/openshift_nightlies/scripts/run_benchmark.sh
whitleykeith: Yeah I looked at the actual code in network-perf and it looks like it *should* do a pip install everytime the script runs. I saw that it wanted me to do that in the readme which is why I added it but that may just be outdated so I'll remove this and see if the uperf workloads still work properly.

amitsagtani97: yes, else we can add the pip install command in https://github.com/cloud-bulldozer/e2e-benchmarking/blob/master/workloads/network-perf/common.sh

amitsagtani97 (approved):
LGTM!

Since the Airflow workflow is completely new to me, I cannot review it in depth, but since these changes are working fine and they are much required updates, let's merge them and we can fix them in case we get some errors in the future.

PR #12, opened 2021-05-13
kubeadmin password , dynamic build_url generation, and updates from baremetal install

whitleykeith:
* Install task now grabs kubeadmin password from orchestration host and mounts it as a secret to allow other tasks to use it such as benchmarks (required for loggng workload)
* Index tasks now dynamic generate the airflow build url instead of it being hardcoded
* made necessary changes to perfscale default `values.yaml` to support baremetal install of the airflow stack


mohit-sheth (approved):
LGTM

smalleni (approved):
LGTM

PR #16, opened 2021-06-10
Install Abstract Base Class and Developer playground

whitleykeith:
Implements base class for install tasks and adds support for developer playground

this pr only changes the cloud installer to use the ABC. A separate PR will need to be added to support baremetal.

relevant dag: http://whitleykeith-airflow.apps.keith-cluster.perfscale.devcluster.openshift.com/graph?dag_id=4.6_aws_default


amitsagtani97 (approved):
LGTM!

PR #19, opened 2021-06-15
Created Shift-on-stack task

amitsagtani97:
### Description
Created shift-on-stack task


dags/openshift_nightlies/dag.py
```diff
@@ -111,14 +112,35 @@ def build(self):
     def _get_openshift_installer(self):
         return jetski.BaremetalOpenshiftInstaller(self.dag, self.version, self.release_stream, self.platform, self.profile, self.openshift_build)
 
-        
+class ShiftOnStackNightlyDAG(AbstractOpenshiftNightlyDAG):
+    def __init__(self, version, release_stream, platform, profile, version_alias, build):
+        super().__init__(version, release_stream, platform, profile, version_alias)
+        self.release_stream_base_url = Variable.get("release_stream_base_url")
+        self.latest_release = var_loader.get_latest_release_from_stream(self.release_stream_base_url, self.release_stream)
```
whitleykeith, line 119: `self.release_stream_base_url` is set to https://openshift-release.apps.ci.l2s4.p1.openshiftapps.com/ so the release_stream passed through the manifest has to be one that exists from this URL. Moreover, `var_loader.get_latest_release_from_stream()` assumes a certain API spec from the `release_stream_base_url`. Setting the release stream to `latest-4.7` won't work with this. See https://github.com/cloud-bulldozer/airflow-kubernetes/blob/master/dags/openshift_nightlies/util/var_loader.py#L18

dags/openshift_nightlies/manifest.yaml
```diff
@@ -58,3 +58,8 @@ platforms:
       build: dev
       profiles:
       - jetski
+  openstack:
+    - version: 4.7
+      releaseStream: latest-4.7
```
whitleykeith, line 63: See comment on `dags/openshift_nightlies/dag.py` L119

dags/openshift_nightlies/releases/4.7/openstack/default/install.json
```diff
@@ -0,0 +1,26 @@
+{
+    "orchestration_user": "stack",
+    "public_net_name": "nova",
+    "osp_project_name": "shiftstack",
+    "openstack_create_flavors": "true",
+    "openstack_upload_image": "true",
+    "openshift_base_domain": "example.com",
```
whitleykeith, line 7: is this default usable? If there's a variable with no default that *needs* to be added in release-specific variables then set them to `""` to indicate that a user should add this. See https://github.com/cloud-bulldozer/airflow-kubernetes/blob/master/dags/openshift_nightlies/tasks/install/cloud/defaults.json

dags/openshift_nightlies/tasks/install/openstack/shift_on_stack.py
```diff
@@ -0,0 +1,30 @@
...
+
+import json
+import requests
+
+from airflow.operators.bash_operator import BashOperator
+from airflow.models import Variable
+from kubernetes.client import models as k8s
+
+# Defines Tasks for installation of Openshift Clusters
+class ShiftOnStackInstaller(AbstractOpenshiftInstaller):
```
whitleykeith, line 17: Can this be called either `openstack.py`/`OpenstackInstaller` or `jetpack.py`/`OpenstackJetpackInstaller`  to make it easier to map the classes to what platform/installer they are responsible for

whitleykeith (changes requested):
See comments

dags/openshift_nightlies/tasks/install/openstack/shift_on_stack.py
amitsagtani97: fixed

dags/openshift_nightlies/dag.py
```diff
@@ -111,14 +112,35 @@ def build(self):
     def _get_openshift_installer(self):
         return jetski.BaremetalOpenshiftInstaller(self.dag, self.version, self.release_stream, self.platform, self.profile, self.openshift_build)
 
-        
+class ShiftOnStackNightlyDAG(AbstractOpenshiftNightlyDAG):
```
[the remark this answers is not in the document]
amitsagtani97: fixed

dags/openshift_nightlies/dag.py
amitsagtani97: removed the stream vars, will be using latest-4.7/4.8

dags/openshift_nightlies/manifest.yaml
amitsagtani97: fixed

dags/openshift_nightlies/releases/4.7/openstack/default/install.json
amitsagtani97: fixed

amitsagtani97:
Opening a new pr from personal fork to use the playground scripts!

PR #24, opened 2021-06-22
shift-on-stack CI

amitsagtani97:


dags/openshift_nightlies/util/manifest.py
```diff
@@ -46,8 +46,24 @@ def get_baremetal_releases(self):
                     "versionAlias": version_alias,
                     "profile": profile
                 })
+            
+    def get_openstack_releases(self):
+        for version in self.yaml['platforms']['openstack']:
```
whitleykeith, line 51: change this to `for version in self.yaml['platforms'].get('openstack', [])` so that manifests with no defined releases for openstack don't error out.

whitleykeith (approved):
had one little comment about the for loop in the manifest class but other than that so long as it works as expected in your env it LGTM

dags/openshift_nightlies/tasks/install/openstack/jetpack.py
```diff
@@ -0,0 +1,61 @@
...
+            trigger_rule=trigger_rule,
+            executor_config=self.exec_config,
+            env=self.env
+        )
+    
+    def _setup_task(self, operation="install"):
+        self.config = {**self.config, **self._get_playbook_operations(operation)}
+        self.config['openshift_cluster_name'] = self.openstack_creds["ocp_cluster_name"]
+        self.config['dynamic_deploy_path'] = self.openstack_creds["ocp_cluster_name"]
+        self.config['kubeconfig_path'] = "home/stack/" + self.openstack_creds["ocp_cluster_name"] + "/auth/kubeconfig"
```
masco, line 36: /home/stack/---- / is missing

amitsagtani97: fixed

dags/openshift_nightlies/util/manifest.py
amitsagtani97: fixed

PR #27, opened 2021-06-29
Adding Bug Fixes from playground tests

whitleykeith:
Fixes some of the bugs people found from using airflow playgrounds:

* Changed the scripts from using the release specific oc client to using a stable one to prevent benchmarks failing when the url isn't resolvable 
* Changed es url logic to support not having username/password 
* Turned off remote_logging with elastic in playgrounds to avoid logging conflicts when multiple airflows run the same scheduled dag
* Removed 4.7 manifests 
* fixed oc login cert issue 
* move to the v1 airflow helm chart
* add links to kibana for better log visibility
* changing some defaults to make the developer environment more dev-like
* adding namespace to log_id to prevent airflow from grabbing logs from different runs


jtaleric (approved):
LGTM

dags/openshift_nightlies/tasks/index/status.py
```diff
@@ -51,10 +31,14 @@ def __init__(self, dag, version, release_stream, latest_release, platform, profi
...
         self.env = {
-            "OPENSHIFT_CLIENT_LOCATION": self.latest_release["openshift_client_location"],
             "RELEASE_STREAM": self.release_stream,
             "TASK": self.task
         }
+        self.git_user = var_loader.get_git_user()
+        if self.git_user == 'cloud-bulldozer'
+            self.env["ES_INDEX"] = "perf_scale_ci"
+        else:
+            self.env["ES_INDEX"] = f"{self.git_user}_playground"
```
mohit-sheth, line 41: when we spin up a playground does it create this index?

whitleykeith: it will the first time a task is indexed

jtaleric (approved):
LGTM!

mohit-sheth (approved):
LGTM

PR #28, opened 2021-07-01
Fix osp vars

amitsagtani97:
### Description
 - orchestration host vars were overriding, renamed var for osp.
 - Disabled multi regions nodes checking for upef in shift-on-stack.
 - Updated schedule interval for shift-on-stack pipeline.


whitleykeith (approved):
lgtm

PR #31, opened 2021-07-02
seperate snappy folder for each playground user

Harshith-umesh:
### Description
Uses github username to create a seperate user folder in snappy , "perf-ci" is the default for the main airflow instance , and a seperate new folder with your github name is created for each playground user. 

@whitleykeith


whitleykeith (approved):
lgtm

PR #32, opened 2021-07-13
Qol updates

whitleykeith:
### Description

Some QOL Updates to make development easier as our dags are getting more complex.

### Highlights

* Parameterizes the dag `schedule_interval` within the manifest at the platform level while allowing for a default. 
* `schedules.enabled` added to the manifest which lets users turn off scheduling and only allow manual triggering of DAGs.
* Creates models for  `OpenshiftRelease`, `BaremetalRelease`, and  `DagConfig` to make handling the release/dag configurations easier throughout the codebase. 
* Made the Baremetal installer inherit from the `AbstractOpenshiftInstaller` 

4.8 test run http://whitleykeith-qol-updates-airflow.apps.sailplane.perf.lab.eng.rdu2.redhat.com/tree?dag_id=4.8_aws_default

### DIFF
--- a/dags/openshift_nightlies/dag.py
+++ b/dags/openshift_nightlies/dag.py
@@ -12,6 +12,8 @@ from airflow.utils.task_group import TaskGroup
 
 # Configure Path to have the Python Module on it
 sys.path.insert(0,os.path.abspath(os.path.dirname(__file__)))
+from models.dag_config import DagConfig
+from models.release import OpenshiftRelease, BaremetalRelease
 from tasks.install.cloud import openshift
 from tasks.install.openstack import jetpack
 from tasks.install.baremetal import jetski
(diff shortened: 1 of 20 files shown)


whitleykeith:
@kedark3

dags/openshift_nightlies/dag.py
```diff
@@ -77,86 +63,78 @@ def _get_openshift_installer(self):
...
         with TaskGroup("benchmarks", prefix_group_id=False, dag=self.dag) as benchmarks:
             benchmark_tasks = self._get_e2e_benchmarks().get_benchmarks()
             chain(*benchmark_tasks)
-            benchmark_tasks[-1] >> utils
-
-        install_cluster >> benchmarks
+            
+        if self.config.cleanup_on_success:
+            cleanup_cluster = installer.get_cleanup_task()
+            install_cluster >> benchmarks >> utils >> cleanup_cluster
```
kedark3, line 87: Is this some kind of ordering you specifying ?

dags/openshift_nightlies/scripts/install/cloud.sh
```diff
@@ -20,7 +20,6 @@ setup(){
     git clone https://${SSHKEY_TOKEN}@github.com/redhat-performance/perf-dept.git
     export PUBLIC_KEY=/home/airflow/workspace/perf-dept/ssh_keys/id_rsa_pbench_ec2.pub
     export PRIVATE_KEY=/home/airflow/workspace/perf-dept/ssh_keys/id_rsa_pbench_ec2 
-    export ANSIBLE_FORCE_COLOR=true
```
kedark3, line 23: why removing this, AirFlow doesn't support it? curious.

dags/openshift_nightlies/util/manifest.py
```diff
@@ -1,69 +1,112 @@
 import yaml
-
+import sys
```
kedark3, line 2: Seems like these imports may be able to use some sorting. 
Maybe use Flake8 or Black linter ?

dags/openshift_nightlies/dag.py
whitleykeith: yeah this is actually drawing the dag dependencies and stuff. Essentially this is saying "cleanup_cluster depends on utils which depends on the benchmarks which depend on install_cluster"

dags/openshift_nightlies/scripts/install/cloud.sh
whitleykeith: yeah airflow doesn't support the colored output and I think it might be the cause of fluentd rendering invalid characters in the logs (at least for the install task).

dags/openshift_nightlies/util/manifest.py
whitleykeith: yeah the imports need a cleanup. The main problem with using a linter atm is that the [PYTHONPATH is a bit messed up in Airflow](https://newbedev.com/apache-airflow-dag-cannot-import-local-module) to where I have to do some sys.path stuff to import local modules:

```python
import sys
from os.path import abspath, dirname
from os import environ

# need to call sys.path.insert before importing
sys.path.insert(0, dirname(dirname(abspath(dirname(__file__)))))
from util import var_loader, kubeconfig, constants
from tasks.index.status import StatusIndexer
from models.release import OpenshiftRelease
```

My VSCode Python extension will move the sys.path.insert call to after the local module imports everytime and it breaks the code. There's probably a better way to solve this though

Harshith-umesh (approved):
LGTM

dags/openshift_nightlies/dag.py
kedark3: Can you please add that as a comment in the code itself? Could make things more readable.

dags/openshift_nightlies/scripts/install/cloud.sh
kedark3: +1 for response, -1 for airflow :laughing:

dags/openshift_nightlies/util/manifest.py
kedark3: Its ok for the time being,  I can try and fix that as I start some contributions.

kedark3 (approved):
LGTM. Lot of great changes, thanks :tada: :taco:

PR #34, opened 2021-07-16
sequential uperf runs instead of parallel

mohit-sheth:
### Fixes
clients/servers of different types(hn,p2p..) of test land on the same node which will impact the results


whitleykeith (approved):
lgtm

PR #35, opened 2021-07-21
Adding ability to accept aws region

kedark3:
Signed-off-by: Kedar Vijay Kulkarni <kkulkarni@redhat.com>

### Description
Be able to specify the region part of the aws creds in the airflow vars.

### Usage
```{ "aws_access_key_id": "key", "aws_profile": "default", "aws_secret_access_key": "another_key", "aws_region_for_openshift": "us-east-2" }```

Without the region, install might fail. This is to force users to be conscious about where they are installing the cluster, and not hit quota or billing issues. Changing from `aws_region` to `aws_region_for_openshift` -> as if we use `aws_region` it fails in the post install here: https://github.com/cloud-bulldozer/scale-ci-deploy/blob/89116d091c9007d1b579c847ffd36f0398228431/OCP-4.X/roles/post-install/templates/aws-infra-node-machineset.yml.j2


whitleykeith (approved):
lgtm

PR #36, opened 2021-07-22
fixing playground user templates to support restricted kubeconfig

whitleykeith:
this along with a restricted kubeconfig will allow people to create necessary objects in our cluster without giving them full admin access.


kedark3 (approved):
LGTM . Thank you @whitleykeith !!

PR #39, opened 2021-07-29
Add DAGs for ROSA and OSDE2E

morenod:
### Description
To integrate ROSA installation, following change on shared code has been required:
- New class RosaNightlyDAG for AbstractOpenshiftNightlyDAG
- New platform: "rosa"
- Rosa installation scripts (rosa.sh & osde2e.sh)
- New variables on defaults.json:
  -  **airflow_executor_image**, to select the image used by the executor, defaulted to current one **quay.io/keithwhitley4/airflow-ansible:2.1.0**
  - **rosa_installation_method**, to select installation on ROSA using rosa cli or osde2e cli
- New secret, **rosa_creds** containing a valid token to access OCM stage env
- Cluster Name for ROSA cannot exceed 15chars, _generate_cluster_name is overrided in the ROSA class to generate a name based on "airflow-<<cluster version without dots (47,48,49,etc)>>-md5(original-cluster-name)[4chars]". This results on a unique name for each combination of githubuser+release+platform, but enclosed to 15chars
- OCP secret related to authentication, kubeadmin password has been set to optional, as they are not generated by default during the rosa installation
- airflow-managedservices Dockerfile generated from the official airflow image, but containing rosa, osde2e and ocm tools
- Using **openshift_network_type** var to install OVN using OSDE2E


dags/openshift_nightlies/scripts/install/rosa.sh
```diff
@@ -0,0 +1,71 @@
...
+    export AWS_SECRET_ACCESS_KEY=$(cat ${json_file} | jq -r .aws_secret_access_key)
+    export ROSA_TOKEN=$(cat ${json_file} | jq -r .rosa_creds_token)
+    export CLUSTER_NAME=$(cat ${json_file} | jq -r .openshift_cluster_name)
+    rosa login --env=staging
+}
+
+install(){
+    rosa whoami
+    rosa verify quota
+    rosa verify permissions
```
dry923, line 32: I assume the above rosa commands are to verify things but shouldn't we be exiting on failure?

dags/openshift_nightlies/scripts/install/rosa.sh
```diff
@@ -0,0 +1,71 @@
...
+setup(){
+    mkdir /home/airflow/workspace
+    cd /home/airflow/workspace
+    export PATH=$PATH:/usr/bin
+    export AWS_REGION=us-west-2
+    export AWS_ACCESS_KEY_ID=$(cat ${json_file} | jq -r .aws_access_key_id)
+    export AWS_SECRET_ACCESS_KEY=$(cat ${json_file} | jq -r .aws_secret_access_key)
+    export ROSA_TOKEN=$(cat ${json_file} | jq -r .rosa_creds_token)
+    export CLUSTER_NAME=$(cat ${json_file} | jq -r .openshift_cluster_name)
+    rosa login --env=staging
```
dry923, line 26: can we make the env string a variable? It would allow us more flexibility in the future

dry923:
@morenod Can we add something to also handle the creation of the workload node? I didn't see it in there atm.

dags/openshift_nightlies/scripts/install/rosa.sh
morenod: if any of them fails, script ends with exit code 1, there is no need to get result.

actually, im going to remove the exit code verification on the rosa create cluster, because it is not being used, thats from an old version with "|| true" after the command

dry923: in that case should we move the verification to the setup section after the login?

dags/openshift_nightlies/scripts/install/rosa.sh
morenod: yep, Im going to add as param. and also save a token also for prod

dags/openshift_nightlies/scripts/install/rosa.sh
morenod: moved

dags/openshift_nightlies/scripts/install/rosa.sh
morenod: added tokens for prod and stage, and make environment a var for each release

morenod:
> @morenod Can we add something to also handle the creation of the workload node? I didn't see it in there atm.

I will do it on another PR

whitleykeith (changes requested):
small nit around kubeconfig otherwise lgtm

dags/openshift_nightlies/util/kubeconfig.py
```diff
@@ -22,7 +23,8 @@ def get_kubeconfig_volume(release: OpenshiftRelease):
     return k8s.V1Volume(
         name="kubeconfig",
         secret=k8s.V1SecretVolumeSource(
-            secret_name=f"{release.version}-{release.platform}-{release.profile}-kubeconfig"
+            secret_name=f"{release.version}-{release.platform}-{release.profile}-kubeconfig",
+            optional=True
```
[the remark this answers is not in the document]
morenod: fixed, only password will be optional

morenod:
@dry923 @whitleykeith please review, I need this change merged to continue with osde2e integration

dags/openshift_nightlies/tasks/install/openshift.py
```diff
@@ -26,18 +25,21 @@ def __init__(self, dag, release: OpenshiftRelease):
...
-
-        self.install_secrets = var_loader.get_secret(
-            f"openshift_install_config", deserialize_json=True)
-        self.aws_creds = var_loader.get_secret("aws_creds", deserialize_json=True)
-        self.gcp_creds = var_loader.get_secret("gcp_creds", deserialize_json=True)
-        self.azure_creds = var_loader.get_secret("azure_creds", deserialize_json=True)
-        self.ocp_pull_secret = var_loader.get_secret("osp_ocp_pull_creds")
-        self.openstack_creds = var_loader.get_secret("openstack_creds", deserialize_json=True)
-        self.release_stream_base_url = var_loader.get_secret("release_stream_base_url")
+        self.ansible_orchestrator = Variable.get("ansible_orchestrator", deserialize_json=True)
```
whitleykeith, line 34: every Variable we grab needs to be grabbed from `var_loader.get_secret`. It takes the same arguments

whitleykeith (changes requested):
@morenod see the comment about grabbing Variables. With Vault we should use the `var_loader.get_secret` function instead of using `Variable.get` to allow for user-defined variable overrides. 

Other than that as long as the DAGs compile (you can turn the DAG on and run it), lgtm

dags/openshift_nightlies/tasks/install/openshift.py
morenod: ok, I ate this commit on the PR, needs to rebase

dags/openshift_nightlies/scripts/install/osde2e.sh
```diff
@@ -0,0 +1,103 @@
...
+cloudProvider:
+  providerId: aws
+  region: ${AWS_REGION}
+cluster:
+  multiAZ: true
+  name: ${CLUSTER_NAME}
+ocm:
+  token: "${ROSA_TOKEN}"
+rosa:
+  awsAccessKey: ${AWS_ACCESS_KEY_ID}
```
dry923, line 53: We also need the ability to set the worker count and worker type. ie:
```
rosa:  
  computeNodes: 30
  computeMachineType: m5.2xlarge
```

dags/openshift_nightlies/scripts/install/osde2e.sh
```diff
@@ -0,0 +1,103 @@
...
+    fi
+    postinstall
+}
+
+postinstall(){
+    _download_kubeconfig $(_get_cluster_id ${CLUSTER_NAME}) ./kubeconfig
+    kubectl delete secret ${KUBECONFIG_NAME} || true
+    kubectl create secret generic ${KUBECONFIG_NAME} --from-file=config=./kubeconfig
+    # set expiration to 24h
+    rosa edit cluster -c $(_get_cluster_id ${CLUSTER_NAME}) --expiration=1440m
```
dry923, line 73: Is it viable to have the cluster expiration a var or do you think that wouldn't be worth the addition?

dags/openshift_nightlies/tasks/install/rosa/defaults.json
```diff
@@ -0,0 +1,63 @@
...
+    "openshift_machine_cidr": "10.0.0.0/16",
+    "openshift_service_network": "172.30.0.0/16",
+    "openshift_host_prefix": "22",
+    "openshift_network_type": "",
+    "openshift_post_install_poll_attempts": "600",
+    "openshift_toggle_infra_node": true,
+    "openshift_toggle_workload_node": true,
+    "dittybopper_enable": true,
+    "kubeconfig_path": "",
+    "cerberus_config_path": "/root/cerberus.yml",
```
dry923, line 29: Cerberus was deprecated out so We shouldn't need these vars anymore.

dags/openshift_nightlies/util/executor.py
```diff
@@ -21,14 +21,14 @@ def get_default_executor_config():
             )
         }
 
-def get_executor_config_with_cluster_access(release: OpenshiftRelease):
+def get_executor_config_with_cluster_access(release: OpenshiftRelease, executor_image='quay.io/keithwhitley4/airflow-ansible:2.1.0'):
```
dry923, line 24: @whitleykeith thoughts on moving your image to the cloud-bulldozer quay org?

dags/openshift_nightlies/releases/4.7/rosa/default/install.json
```diff
@@ -0,0 +1,32 @@
+{
```
dry923, line 1: Much of these defaults have no use in rosa. I.e. we dont change the master instance type, etc. Can we remove the unneeded ones or will that break things?

dags/openshift_nightlies/scripts/install/osde2e.sh
morenod: It is, but we have to think on this job as a cycle job, creates > benchmark > destroy. We can calculate the required expiration or we can make it as variable, but it makes no sense to have an expiration higher tan the time required for the benchmak, because the cluster will be automatically destroyed after tests

dags/openshift_nightlies/scripts/install/osde2e.sh
morenod: Added

dags/openshift_nightlies/tasks/install/rosa/defaults.json
morenod: I have just copied the defaults file from the cloud platform.
Im not using almost any of them during the installation of rosa, but Im not sure to remove them, as I dont know if they are required during the benchmark execution.

@whitleykeith can you confirm that if I remove any non used var from the installation defaults.json, it wont break anything?

dags/openshift_nightlies/releases/4.7/rosa/default/install.json
morenod: same than https://github.com/cloud-bulldozer/airflow-kubernetes/pull/39#discussion_r693764207

I cannot confirm that deleting those vars will affect or not the benchmarking

dags/openshift_nightlies/tasks/install/rosa/defaults.json
whitleykeith: yeah you can remove them. The install vars are only for install and your defaults.json is only for your specific install class

dags/openshift_nightlies/util/executor.py
whitleykeith: I'm working on [this branch](https://github.com/whitleykeith/airflow-kubernetes/tree/ci) to add a CI framework to Airflow which would include doing that

dags/openshift_nightlies/scripts/install/osde2e.sh
morenod: added

dags/openshift_nightlies/tasks/install/rosa/defaults.json
morenod: all removed

dags/openshift_nightlies/releases/4.7/rosa/default/install.json
morenod: all removed

whitleykeith (approved):

dry923 (approved):
LGTM!

PR #41, opened 2021-08-06
Baremetal DAGs with benchmark

mukrishn:
### Description

1. Includes changes for Airflow 2.1.0
2. DAGs for OCP 4.8 changes 
3. Includes benchmark tasks 
4. and more cleanups


dags/openshift_nightlies/tasks/benchmarks/e2e.py
```diff
@@ -76,13 +94,13 @@ def _add_indexers(self, benchmarks):
...
     def _add_indexer(self, benchmark): 
-        indexer = StatusIndexer(self.dag, self.config, self.release, benchmark.task_id).get_index_task() 
+        indexer = StatusIndexer(self.dag, self.dag_config, self.release, benchmark.task_id).get_index_task() 
         benchmark >> indexer 
 
     def _get_benchmark(self, benchmark):
         env = {**self.env, **benchmark.get('env', {}), **{"ES_SERVER": var_loader.get_secret('elasticsearch')}, **{"KUBEADMIN_PASSWORD": environ.get("KUBEADMIN_PASSWORD", "")}}
         return BashOperator(
-            task_id=f"{benchmark['name']}",
+            task_id=f"{self.task_group}_{benchmark['name']}",
```
whitleykeith, line 103: can you make this conditional to where we don't prefix the names unless the task_group is not the default `benchmarks`?

whitleykeith (changes requested):

dags/openshift_nightlies/tasks/benchmarks/e2e.py
mukrishn: Yes, makes sense, included a conditional prefix.

whitleykeith (approved):

PR #42, opened 2021-08-09
Enable thanos for all cloud deployments

whitleykeith:
* turns on thanos remote_write in cloud installs
* fixes issue in es operator application where argo gets stuck watching a field
* adds make to the airflow executor image

depends on https://github.com/cloud-bulldozer/scale-ci-deploy/pull/161


mohit-sheth (approved):
LGTM

PR #45, opened 2021-08-24
Airflow CI + Chart Migration

whitleykeith:
### Description

* Move remote_logging for Airflow tasks from elasticsearch to s3 to fix issues with logs being out of order due to upstream bugs 
* Remove local copy of Airflow chart in favor of using versioned upstream one from Apache
* adds testing framework and core tests against DAG code
* adds multi-stage container builds for an Airflow core + executor images 
* Github Actions for DAG Unit Testing to run on PRs
* Github Action for pushing Airflow Images on pushes to master
* `executorImages` section added to `dagConfig` in the manifest that lets users define which quay org/tag they want to use for their Airflow tasks
* some blocks on the perfscale install script added to prevent people from running that script by mistake.


dags/openshift_nightlies/tasks/install/rosa/rosa.py
```diff
@@ -19,6 +19,9 @@
 # Defines Tasks for installation of Openshift Clusters
 
 class RosaInstaller(AbstractOpenshiftInstaller):
+    def __init__(self, dag, config: DagConfig, release: OpenshiftRelease):
+        super().__init__(dag, config, release)
+        self.exec_config = self.exec_config = executor.get_default_executor_config(self.dag_config, executor_image="airflow-managed-services")
```
morenod, line 24: is this sentence ok? I think there is an extra self.exec_config

morenod (changes requested):

dags/openshift_nightlies/tasks/install/rosa/rosa.py
whitleykeith: yep, good catch. Fixed

jtaleric (approved):
lgtm - small nit/question -- do we want to keep compare enabled for uperf. Or disable even though we moved it. @mohit-sheth for last signoff.

mohit-sheth (approved):
LGTM

PR #52, opened 2021-09-07
fix ocp 4.9 on osp

masco:


amitsagtani97 (changes requested):
Shall we remove this var from 4.7 and 4.8 version as well?

amitsagtani97 (approved):

PR #53, opened 2021-09-07
add ocp_network_type var from vault

amitsagtani97:
### Description
Get network type var from vault to override in case we need to run test on a particular network type in shift-on-stack CI

### DIFF
--- a/dags/openshift_nightlies/tasks/install/openstack/jetpack.py
+++ b/dags/openshift_nightlies/tasks/install/openstack/jetpack.py
@@ -43,6 +43,11 @@ class OpenstackJetpackInstaller(AbstractOpenshiftInstaller):
             **self._insert_kube_env()
         }
 
+        # Here if network type is defined in the vars, it will be set else, network type
+        # is decided based on the day of the week.
+        if self.config['ocp_network_type'] != "":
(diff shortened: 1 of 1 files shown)


whitleykeith:
why do you need to add this logic via secret instead of just making a profile for each network type, similar to how we do it for the cloud deployments?

masco:
@whitleykeith currently shift on stack is running on a scale lab LTA allocation and we have just one deployment of OSP. we can't run the jobs parallelly on it. but still, we need to test all the three SDNs(Kuryr, OpenshiftSDN, and OVNKubernetes). so we thought choosing the SDN using the days logic is easier comparatively having three jobs and synchronizing between them. The SDN is chosen automatically based on the date for any job triggered by a cron job. this vault variable is used only if we want to test any particular SDN manually.

jtaleric (changes requested):
I don't think this makes sense to do a CNI based on the day it ran... the nightly builds could be different, so why would you compare them? 

I would prefer we focus on a single CNI.

amitsagtani97:
@jtaleric makes sense! Removed the day based logic and set specific CNI for a particular DAG.
Closing this PR.

PR #55, opened 2021-09-14
do not use nightlies for 4.7 and 4.8 OSP jobs

masco:
### Description
4.7 and 4.8 jobs taking nightly build, since there are no nightlies on these versions removing that.


amitsagtani97 (approved):
LGTM!

PR #57, opened 2021-09-16
disable debug option on workload node for OSP jobs

masco:

### DIFF
--- a/dags/openshift_nightlies/releases/4.7/openstack/default/install.json
+++ b/dags/openshift_nightlies/releases/4.7/openstack/default/install.json
@@ -54,5 +54,6 @@
     "openshift_prometheus_storage_size": "10Gi",
     "openshift_alertmanager_storage_class": "standard",
     "openshift_alertmanager_storage_size": "2Gi",
+    "openshift_toggle_workload_node": "false",
     "job_iterations": "1"
  }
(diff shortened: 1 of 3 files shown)


dags/openshift_nightlies/scripts/install/jetpack.sh
```diff
@@ -66,7 +66,7 @@ run_jetpack(){
         # disable openshift install as we already installed via jetpack
         export OPENSHIFT_INSTALL=false
         export DYNAMIC_DEPLOY_PATH=$DEPLOY_PATH
-        time /home/airflow/.local/bin/ansible-playbook -i ../inventory -vv OCP-4.X/install-on-osp.yml --extra-vars "@${json_file}" | tee $(date +"%Y%m%d-%H%M%S")-post-install.timing
+        time /home/airflow/.local/bin/ansible-playbook -i ../inventory -vv OCP-4.X/install-on-osp.yml -e openshift_toggle_workload_node=false --extra-vars "@${json_file}" | tee $(date +"%Y%m%d-%H%M%S")-post-install.timing
```
amitsagtani97, line 69: @masco why pass this var individually? We can pass it with other --extra-vars in install.json

amitsagtani97 (changes requested):

dags/openshift_nightlies/scripts/install/jetpack.sh
masco: we need this var as `true` on jetpack, since we are creating the workload nodes from jetpack.

amitsagtani97 (approved):

PR #58, opened 2021-09-16
Disable the ACS install task for cleanup

gavin-stackrox:
Ensures that when `rhacs_enable` is set it is overridden for the cleanup task. i.e. so that there is no attempt to install ACS during cleanup.


jtaleric (approved):
LGTM

PR #62, opened 2021-09-30
add node-density heavy to the perfscale CI

amitsagtani97:
### Description
Added node-density heavy at 25 scale and 250 pods per node config, we mention results for this config in our reports and can be helpful to track in our CI.

### DIFF
--- a/dags/openshift_nightlies/releases/4.7/openstack/default/benchmarks.json
+++ b/dags/openshift_nightlies/releases/4.7/openstack/default/benchmarks.json
@@ -51,6 +51,24 @@
             }
         },
         {
+            "name": "node_density_heavy",
+            "workload": "kube-burner",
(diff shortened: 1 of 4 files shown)


whitleykeith:
@amitsagtani97 do we still want to do this or are we removing the 4.7/4.8 variants?

jtaleric:
Right we want to remove 4.7 from our view.

amitsagtani97:
@whitleykeith Yes, we will remove the 4.7/4.8. This pr will add node-density-heavy to all the default profiles.
I think we can merge it before removing 4.7/4.8

Harshith-umesh:
closing since this has been incorporated in the manifest v3 pr : https://github.com/cloud-bulldozer/airflow-kubernetes/pull/83

PR #64, opened 2021-10-01
Manual UUID

whitleykeith:
Generates benchmark UUID outside of e2e scripts in order to pass them to indexers to make it easier to find benchmark uuids

Depends on https://github.com/cloud-bulldozer/e2e-benchmarking/pull/256

### DIFF
--- a/dags/openshift_nightlies/scripts/run_benchmark.sh
+++ b/dags/openshift_nightlies/scripts/run_benchmark.sh
@@ -77,6 +77,6 @@ else
     eval "$command"
     benchmark_rv=$?
     echo $UUID
-    exit benchmark_rv
+    exit $benchmark_rv
 
 fi
--- a/dags/openshift_nightlies/scripts/run_benchmark.sh
+++ b/dags/openshift_nightlies/scripts/run_benchmark.sh
@@ -14,7 +14,7 @@ done
 setup(){
     mkdir /home/airflow/workspace
     cd /home/airflow/workspace
-    git clone https://github.com/cloud-bulldozer/e2e-benchmarking
+    git clone -b manual-uuid https://github.com/whitleykeith/e2e-benchmarking
     cp /home/airflow/auth/config /home/airflow/workspace/config
     export KUBECONFIG=/home/airflow/workspace/config
     curl http://dell-r510-01.perf.lab.eng.rdu2.redhat.com/msheth/gsheet_key.json > /tmp/key.json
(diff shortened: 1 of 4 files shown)


dags/openshift_nightlies/scripts/run_benchmark.sh
```diff
@@ -14,7 +11,7 @@ done
 setup(){
     mkdir /home/airflow/workspace
     cd /home/airflow/workspace
-    git clone https://github.com/cloud-bulldozer/e2e-benchmarking
+    git clone -b manual-uuid https://github.com/whitleykeith/e2e-benchmarking
```
mohit-sheth, line 14: merged the upstream e2e PR

dags/openshift_nightlies/scripts/run_benchmark.sh
```diff
@@ -68,13 +65,18 @@ EOF
 }
 
 if [[ $PLATFORM == "baremetal" ]]; then
-env >> /tmp/environment.txt
-run_baremetal_benchmark
-echo "Finished e2e scripts for $workload"
+    env >> /tmp/environment.txt
```
mohit-sheth, line 68: tabs?

whitleykeith: yeah just for readability

dags/openshift_nightlies/scripts/run_benchmark.sh
```diff
@@ -14,7 +11,7 @@ done
 setup(){
     mkdir /home/airflow/workspace
     cd /home/airflow/workspace
-    git clone https://github.com/cloud-bulldozer/e2e-benchmarking
+    git clone -b master https://github.com/cloud-bulldozer/e2e-benchmarking
```
sjug, line 14: Isn't the default branch `master`?

sjug (approved):
Small nit, lgtm otherwise.

dags/openshift_nightlies/scripts/run_benchmark.sh
jtaleric: If we want to call out the branch, it might make sense to make into a env var we can override. Default it to `master`.

whitleykeith: yeah i did that because I was gonna work on adding a dependency section in the dag config so we can change it there instead of having to mess with the scripts. I was just gonna do that in a separate PR so it didn't block this

Also I tend to do this pretty frequently and this way I can just replace `cloud-bulldozer`

jtaleric (approved):
LGTM

PR #67, opened 2021-10-08
Configurable e2e-benchmarking tag

rsevilla87:

### Description

Configurable e2e-benchmarking tag. I also updated get_secret to prevent affecting current DAGs

### DIFF
--- a/dags/openshift_nightlies/manifest.yaml
+++ b/dags/openshift_nightlies/manifest.yaml
@@ -19,7 +19,7 @@ dagConfig:
     tag: 2.1.3
   dependencies:
     e2e_benchmarking:
-      repo: git@github.com:cloud-bulldozer/e2e-benchmarking.git
+      repo: https://github.com/cloud-bulldozer/e2e-benchmarking.git
       branch: master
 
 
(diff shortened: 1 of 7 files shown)


dags/openshift_nightlies/tasks/install/cloud/defaults.json
```diff
@@ -58,5 +58,6 @@
     "kubeconfig_auth_dir_path": "",
     "fips": false,
     "thanos_enable": true,
-    "loki_enable": true
-}
\ No newline at end of file
+    "loki_enable": true,
+	"e2e_benchmarking_tag": "master"
```
whitleykeith, line 62: doesnt need to be in this file or any of the install files

whitleykeith (changes requested):
the e2e benchmarking tag should be added via the `manifest.yaml` and read into the `DagConfig` object that gets generated and passed to each task. Something like:


```yaml
dependencies:
  e2e:
    repo: string
    branch: string
  
```

That way we can configure the dependency versions without having to update secrets.

morenod (approved):
LGTM

PR #78, opened 2021-10-27
enable Kuryr SDN on 4.10

masco:


dags/openshift_nightlies/manifest.yaml
```diff
@@ -97,6 +97,7 @@ platforms:
       releaseStream: 4.10.0-0.nightly
       profiles:
       - default
```
Harshith-umesh, line 99: can we rename "default" to openshiftsdn or just sdn? will make it easier to interpret

masco: I see this is the practice we have used in all other clouds

Harshith-umesh: yes but with this new pr: https://github.com/cloud-bulldozer/airflow-kubernetes/pull/79 we're planning on getting rid of the default dag type for aws and azure. when it comes to openstack dags it make even more sense to rename defeault to sdn.
 also with this pr, we can enable different cron schedules for different profiles, do you wan't me to do that for openstack as well?  https://github.com/Harshith-umesh/airflow-kubernetes/blob/b6336e3c014a9526186bafc21a2722614d2ab7dd/dags/openshift_nightlies/manifest.yaml#L16

masco: done

Harshith-umesh (approved):
LGTM

dags/openshift_nightlies/manifest.yaml
whitleykeith: @masco to keep in line with the above PR just use `sdn` instead of `openshift-sdn`

masco: @whitleykeith done

PR #79, opened 2021-10-27
divide airflow default config dags into control plane and data plane dags

Harshith-umesh:
### Description
This PR creates airflow dags according to the requirements mentioned in the doc below:
https://docs.google.com/document/d/19qteIrYyewlITnR5Jiv_Xbq3Kr8Xdrl5mWof7x1dLik/edit#heading=h.yyte2g39nnex 

To add new control or data plane dags, change only needs to be made to the manifest.yaml. Cron schedules for different profiles are enabled. cron schedule for the large dags are set to run on the 15th of every month.
4 new benchmark config files are used by the different dags. https://github.com/Harshith-umesh/airflow-kubernetes/tree/what/dags/openshift_nightlies/tasks/benchmarks 

### Fixes
can close this PR : https://github.com/cloud-bulldozer/airflow-kubernetes/pull/62 since node-density-heavy is added here.


dags/openshift_nightlies/manifest.yaml
```diff
@@ -57,10 +65,21 @@ platforms:
       - name: aws
         profiles:
         - default
-        - ovn
+        - sdn-common-control-plane
```
jtaleric, line 68: I thought we were going to do sdn-common-[cp|dp].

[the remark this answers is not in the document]
jtaleric: IMHO removing this only removes the historical data. I don't think it is worth preserving, since all the results are in ES already. Thoughts?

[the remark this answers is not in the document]
Harshith-umesh: ack, removing it

Harshith-umesh: done

dags/openshift_nightlies/manifest.yaml
```diff
@@ -56,11 +64,20 @@ platforms:
       providers:
       - name: aws
         profiles:
-        - default
-        - ovn
+        - sdn-common-cp
```
whitleykeith, line 67: can we just remove `common` from all the profiles? When combined with the version/platform the dag IDs become long and annoying in visualization tools. I'm more in favor of using those characters for spelling out control/data plane 

Also i don't think it needs to be a part of this PR necessarily but with the profiles using hypens the dag IDs are gonna look something like `4.9_aws_sdn-common-dp` or `4.9_aws_sdn-data-plane` which bugs my ocd

dags/openshift_nightlies/tasks/benchmarks/e2e.py
```diff
@@ -29,8 +29,24 @@ def __init__(self, dag, config: DagConfig, release: OpenshiftRelease, task_group
         self.es_server_baseline = var_loader.get_secret("es_server_baseline")
 
         # Specific Task Configuration
-        self.vars = var_loader.build_task_vars(
-            release=self.release, task=self.task_group)
+
```
whitleykeith, line 32: this seems very cumbersome to manage especially if we add new config_files for different tasks. It'd be more maintainable if we just expanded the manifest to let each profile set it's config_file

dags/openshift_nightlies/util/manifest.py
```diff
@@ -128,7 +129,7 @@ def _get_dependencies(self):
 
     def _get_schedule_for_platform(self, platform):
         schedules = self.yaml['dagConfig']['schedules']
-        if bool(schedules.get("enabled", False) and var_loader.get_git_user() == "cloud-bulldozer"):
+        if bool(schedules.get("enabled", False)): # and var_loader.get_git_user() == "cloud-bulldozer"):
```
whitleykeith, line 132: this needs to be re-added so that playgrounds don't run on a schedule by default

whitleykeith (changes requested):

dags/openshift_nightlies/manifest.yaml
Harshith-umesh: sure. thoughts? @jtaleric

dags/openshift_nightlies/tasks/benchmarks/e2e.py
Harshith-umesh: ahh makes sense

dags/openshift_nightlies/util/manifest.py
Harshith-umesh: yes, will enable it before merging

whitleykeith:
We might want to mess with the structure of the `manifest.yaml` and `releases` directory to make some of these changes easier to implement. As it stands the `releases` directory ends up housing a lot of duplicate data and can be a bit confusing to manage.

I've been thinking about having the manifest look something like this:

```yaml
cloud:
 - version: 4.8
   releaseStream: 4.8.0-0.nightly
   variants:
   - name: aws-sdn
     provider: aws
     schedule: "0 12 * * 1,3,5"
     config:
       install: aws/sdn.json
       benchmarks: common.json
```    

This would let us normalize things like install configs across versions (which we currently do by just having duplicate install configs for each version). It also renames the concept of "profiles" as a "variant" since I've heard most other teams refer to this abstraction that way. 

Thoughts? @jtaleric @Harshith-umesh @mohit-sheth

Harshith-umesh:
> ```yaml
> schedule: "0 12 * * 1,3,5"
>      config:
>        install: aws/sdn.json
>        benchmarks: common.json
> ```

sounds good , what about something like this:
platforms:
  cloud:
    - version: 4.10
      releaseStream: 4.10.0-0.nightly
      providers:
      - name: aws
        variants:
          sdn-common-cp:
	    schedule: "0 12 * * 1,3,5"
     	    config:
       		install: aws/sdn.json
       		benchmarks: common.json
	  sdn-medium:
        
       this will be more inline with our existing code.thoughts?

whitleykeith:
> > ```yaml
> > schedule: "0 12 * * 1,3,5"
> >      config:
> >        install: aws/sdn.json
> >        benchmarks: common.json
> > ```
> 
> sounds good , what about something like this:
> 
> ```
> platforms:
>   cloud:
>     - version: 4.10
>       releaseStream: 4.10.0-0.nightly
>       providers:
>       - name: aws
>         variants:
>           sdn-common-cp:
> 	    schedule: "0 12 * * 1,3,5"
>      	    config:
>        		install: aws/sdn.json
>        		benchmarks: common.json
> 	  sdn-medium:
> ```
> 
> ```
>    this will be more inline with our existing code.thoughts?
> ```

I'm not too worried about keeping it inline with the existing code bc these changes would create a lot of unecessary nesting (specifically with the provider section and the variants being a key) in the manifest. I don't think this has to be a part of this PR though I've been working on this change here https://github.com/whitleykeith/airflow-kubernetes/tree/manifest-v3

Harshith-umesh:
closing since all the changes suggested in this PR have been assimilated in the manifest v3 PR https://github.com/cloud-bulldozer/airflow-kubernetes/pull/83

PR #81, opened 2021-10-29
move _generate_cluster_name function from OpenshiftInstaller class to Release

morenod:
At his moment, function _generate_cluster_name() is located on the openshift installer class, so it cannot be used outside the installation step.

From rosa, we need to invoke this function from the benchmarks, because there is no other way to obtain the cluster name used to install the cluster

This PR moves this function from the openshift installer class to the release class, so it can be used on any step, just importing the release class (which is already imported on benckmarks at this moment).

Also cluster name generated by default it too long for ROSA installation, so based on the platform, now we can control, based on the platform, the cluster name generated.

Test for rosa: http://morenod-cluster-name-fix-airflow.apps.sailplane.perf.lab.eng.rdu2.redhat.com/log?dag_id=4.9_rosa_default&task_id=install&execution_date=2021-10-29T13%3A14%3A22.367933%2B00%3A00

Test for AWS: http://morenod-cluster-name-fix-airflow.apps.sailplane.perf.lab.eng.rdu2.redhat.com/log?dag_id=4.8_aws_default&task_id=install&execution_date=2021-10-29T13%3A55%3A20.750966%2B00%3A00

### DIFF
--- a/dags/openshift_nightlies/models/release.py
+++ b/dags/openshift_nightlies/models/release.py
@@ -1,6 +1,8 @@
 import requests
 from dataclasses import dataclass
 from typing import Optional
+from os import environ
+from hashlib import md5
 
 @dataclass
 class OpenshiftRelease:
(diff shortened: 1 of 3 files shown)


whitleykeith (approved):
lgtm

PR #85, opened 2021-11-15
remove old stuff and fix dictionary generation

whitleykeith:
Fixes and issue where the install configs don't get properly applied because of how python handles dictionary updates

Removes old charts/deploy scripts


jtaleric (approved):
LGTM

PR #89, opened 2021-11-30
Increase ROSA clusters expiration to 2 days

morenod:
ROSA clusters are automatically deleted after some time.

Increasing that time to 2 days, because it seems that 1 day is not enough


chentex (approved):
LGTM

PR #95, opened 2022-01-03
Update cluster name for ROSA as 4.10 exceeds max chars

morenod:
On 4.10 installations, cluster name airflow-410-53f6 exceeds cluster max number of chars.

Modified to perf-410-XXXX so we have some free chars


whitleykeith (approved):
Lgtm

PR #97, opened 2022-01-12
Update UUID gen expression

rsevilla87:
The first part of the UUID is now the results of uuidgen instead of the task name. Using the task name is problematic since for example, kube-burner jobs name are generated with the expression:
`name: kube-burner-{{ trunc_uuid }}`

This trunc_uuid is similar in the node-density and node-density-heavy
benchmarks. (Jobs created here are named kube-burner-node). Running both bencharks in a row It's not possible, since we can't create two jobs with the same name.

### Fixes

Node-density-heavy benchmark can't run after running node-density.

### DIFF
--- a/dags/openshift_nightlies/scripts/run_benchmark.sh
+++ b/dags/openshift_nightlies/scripts/run_benchmark.sh
@@ -83,7 +83,7 @@ else
     cd /home/airflow/workspace
     ls
     cd e2e-benchmarking/workloads/$workload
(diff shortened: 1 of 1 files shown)


morenod (approved):
LGTM

morenod:
Verified working fine at:

http://morenod-uuidgen-airflow.apps.sailplane.perf.lab.eng.rdu2.redhat.com/log?dag_id=4.10-rosa-sdn-control-plane&task_id=node-density-heavy&execution_date=2022-01-13T09%3A11%3A20.234227%2B00%3A00

whitleykeith (approved):
Lgtm

PR #98, opened 2022-01-12
change azure disk sizes

mohit-sheth:
### Description
Azure disks are currently set to 1Tb. This was due to previous speed scaling issues where the disk needed to be that large in order to get the appropriate performance. We can scale this size back down now. The managed services azure workers use 128 for example

### Fixes
https://github.com/cloud-bulldozer/scale-ci-deploy/issues/182


dry923 (approved):
LGTM!

sjug (approved):
/lgtm

PR #100, opened 2022-01-17
Support Airflow Chart v1.4 + owner fix

whitleykeith:
### Description

* Change default values for playgrounds to support deploying off new Version of Airlow Chart as it adds Argo specific functionality 

### Fixes

* DAG Owner Reference to ensure all DAGs are visible in the UI


scripts/playground/templates/airflow.yaml
```diff
@@ -62,13 +70,6 @@ spec:
           replicas: 1
           defaultUser:
             password: $password
-        migrateDatabaseJob:
```
sjug, line 65: Do we not need this `migrateDatabaseJob` anymore?

sjug (commented):
The rest of the DAGs do seem to appear

scripts/playground/templates/airflow.yaml
whitleykeith: we do but the we do not need to mess with the job annotations directly anymore since the upstream chart added the `migrateDatabaseJob.useHelmHooks` field.

whitleykeith: I moved up the block referencing the new value to the top because it's actually very important we disable helm hooks since we deploy via Argo (Argo treats helm hooks slightly differently than helm does and causes initial installs to hang)

smalleni:
@whitleykeith Can we get this merged please?

sjug:
@smalleni Usually the PR owner doesn't merge in their own work, one of the reviewers (usually second) would merge.

smalleni:
> @smalleni Usually the PR owner doesn't merge in their own work, one of the reviewers (usually second) would merge.

Ofcourse :) I didn't want to ping you and the others are not in this timezone ;) 

Anyway, I'm aware of how things need to operate ideally - I was just trying to get folks like you unblocked since you mentioned you had issues getting all DAGs to show up.

PR #104, opened 2022-01-21
Add clustername as DAG Tag

morenod:
Cluster Names are not descriptives when working with ROSA/ARO/ROGCP, because default cluster name exceds the char limit, and we need to create cluster with names as perf-410-b3ec or perf-49-ad41.

This commit adds this cluster name as tag, so we can easily identify the cluster created on each DAG:

image


rsevilla87 (approved):
LGTM

PR #106, opened 2022-01-21
Reduce workload instance sizes to 8x32:

rsevilla87:
### Description

Reduce workload instance sizes in all platorms (Managed services and ACS included cc: @dry923  and @gavin-stackrox ). We're wasting resources & money in beefy workload instances:

New sizes have 8 vCPUs and 32GiB RAM which should be more than enough to run benchmark-operator and kube-burner together:

AWS: m5.2xlarge
Azure: Standard_D8s_v3 
GCP: n1-standard-8


rsevilla87:
@mohit-sheth, I totally forgot about the ingress test. Se should double-check if we hit a client side bottle neck before transitioning to the new instance types.

mohit-sheth:
@rsevilla87 yes
@sjug have you looked at the client resource usage recently?

PR #103, opened 2022-01-20
Rosa ovn

smalleni:
### Description
Enables consumption of the --network-type flag on the CI for deploying OVN clusters in ROSA>

PR #108, opened 2022-01-21
Enable capability to consume ROSA master branch

smalleni:
Right now the cli option in ROSA to deploy OVN clusters is not
in a tagged build. In the future, this might be the case for other
features. It would be nice to have the ability to consume master branch
ROSA. This commit involves changes in the Dockerfile for the
airflow-managed-services image because we would need go in the image and
right now with multi-stage builds we are only using the golang image for
building osde2e binary and then copying  it off to the airflow base
image.

Signed-off-by: Sai Sindhur Malleni <smalleni@redhat.com>


smalleni:
@dry923 @morenod @chentex PTAL. This is needed to build OVN clusters currently but even in the future I'd like to see this capability to keep up with development pace.

dags/openshift_nightlies/config/install/rosa/ovn.json
```diff
@@ -3,7 +3,8 @@
     "aws_access_key_id": "",
     "aws_secret_access_key": "",
     "rosa_environment": "staging",
-    "rosa_installation_method": "osde2e",
+    "rosa_installation_method": "rosa",
+    "rosa_cli_version": "latest",
```
morenod, line 7: Maybe master instead of latest?

smalleni: Intentionally leaving it at latest - we will only use master if need be,

morenod: ok, understood

morenod (approved):
LGTM

smalleni:
Let's get https://github.com/cloud-bulldozer/airflow-kubernetes/pull/103 in first and then I can rebase this PR against that before merging this @morenod @dry923

images/airflow-managed-services/Dockerfile
```diff
@@ -22,7 +25,8 @@ RUN curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2
 RUN unzip awscliv2.zip
 RUN ./aws/install
 RUN curl -sL https://aka.ms/InstallAzureCLIDeb | bash
-COPY --from=build /osde2e/out/osde2e* /usr/local/bin/
+RUN cp /osde2e/out/osde2e* /usr/local/bin/
```
dry923, line 28: Do we need this in here anymore? Rosa OVN was the only thing that was using osde2e that I can recall.

smalleni: Happy to nuke it. I just don't have enough context :)

dry923: @morenod can you confirm we don't use osde2e for any of the other jobs (I'm 99% sure of that) and let's get it out of here its just taking up space.

smalleni: If there's other places where osde2e is referenced, there probably has to be a different PR to remove it though.

dags/openshift_nightlies/config/install/rosa/ovn.json
dry923: @smalleni lets swap the default to master until the new bits are rolled for rosa. Otherwise things will be broken since we merged 103 which looks for that network flag.

images/airflow-managed-services/Dockerfile
morenod: osde2e is only used for OVN installation. I have reviewed that it is removed from everywhere on the PR

smalleni: Ack, thanks!

dags/openshift_nightlies/config/install/rosa/ovn.json
smalleni: ack

dry923 (approved):
LGTM!

morenod (approved):
LGTM

dry923 (approved):
LGTM!

PR #112, opened 2022-01-27
Platform Connector Task

whitleykeith:
### Description

* Creates Platform Connector Task which connects any created cluster to the PerfScale observability components (Thanos, Loki) and generates necessary external labels for querying that data. 
* Adds DAG ID as an external label to all metrics/logs 
* Adds Platform Connector Task to all DAGs, enabling logs + metrics for all existing DAGs
* Deprecates usage of existing post-install tasks from scale-ci-deploy for cloud installs
* Adds Helm3 to all Airflow Images (needed for Platform Connector Task)

### Fixes

* Old Release Logic was making a lot of unecessary calls to the upstream OpenShift Release API. It was calling for every single unique release/DAG and we currently have 70+ DAGs. Changed the logic to happen only once at the beginning of the DAG Import to grab all the new releases for all present versions and pass them through the releases directly.


rsevilla87 (approved):
LGTM!

mukrishn (approved):
Tested on Baremetal and updated code. cc: @mkarg75  @dustinblack

Harshith-umesh:
do we want to enable loki for all the dags, this might cause it to fill up and fall over. also will this also affect playgrounds?

whitleykeith:
> do we want to enable loki for all the dags, this might cause it to fill up and fall over. also will this also affect playgrounds?

yeah all DAGs will send metrics/logs, even playgrounds. For Loki I think it should be fine and we can edit the Loki configs to handle the data if it becomes problematic

PR #113, opened 2022-01-27
use 16x128 instance for master and 16x64 for infra

mohit-sheth:
### Description
Master nodes will have 16vcpu and 1286GB of memory and Infra node will have 16vcpu and 64GB of memory for AWS, Azure and GCP

https://github.com/cloud-bulldozer/airflow-kubernetes/pull/112 can reduce some resource consumption on the Infra nodes, hence not using memory optimized instances for it


rsevilla87 (approved):
Awesome, this will help us to compare apples to apples
LGTM

dry923:
This looks good but it will invalidate a bunch of our previous tests. We may want to wait to do this until we also swap to the m6i, etc type instances as well so we can retrest the bunch at the same time. Thoughts @rsevilla87 @mohit-sheth ?

whitleykeith (approved):
lgtm

PR #117, opened 2022-02-07
Patch adds support to execute Chaos (kraken) tests.

yogananth-subramanian:
Chaos tests are run as an extension of e2e benchmark script by modifying
dags/openshift_nightlies/scripts/run_benchmark.sh script and it uses the
secret override feature to overwrite the aws access details, orchestration
 hosts and sshkeys.The patch introduces the following changes.

- dags/openshift_nightlies/scripts/run_benchmark.sh
  When the workload is set as “kraken“ in benchmark config,
  dags/openshift_nightlies/config/benchmarks/chaos.json, then it triggeres
  the execution of Chaos tests through ansible orchestration host,
  CERBERUS_HOST.

- dags/openshift_nightlies/tasks/benchmarks/e2e.py
  Expose sshkey_token and CERBERUS_HOST, ansible orchestration host, as
  environment variables while executing Openshift benchmark tasks on AWS.

- dags/openshift_nightlies/tasks/install/openshift.py
  Disable triggering ‘cerberus install’ in the playbook during cleanup.

- dags/openshift_nightlies/config/benchmarks/chaos.json
  Chaos test config file.

- dags/openshift_nightlies/config/install/aws/chaos.json
  dags/openshift_nightlies/config/install/azure/chaos.json
  dags/openshift_nightlies/config/install/gcp/chaos.json
  Openshift install config file for chaos testing.


dags/openshift_nightlies/tasks/install/openshift.py.orig
```diff
@@ -0,0 +1,114 @@
+import abc
```
Harshith-umesh, line 1: i think you meant to delete the .orig files.

dags/openshift_nightlies/manifest.yaml
```diff
@@ -75,6 +75,10 @@ platforms:
       config: 
         install: acs.json
         benchmarks: acs.json
+    - name: chaos
+      config:
```
Harshith-umesh, line 79: if you want to create 3 dags one for aws,azure,gcp then you need to do it like how we've done it in the cloud platform, have a providers section and then the variants : 
versions: ["4.10"]
  providers: ["aws", "gcp", "azure"]
    variants: 
    - name: chaos
      config: 
        install: chaos.json
        benchmarks: chaos.json

dags/openshift_nightlies/tasks/benchmarks/e2e.py
```diff
@@ -61,6 +61,21 @@ def __init__(self, dag, config: DagConfig, release: OpenshiftRelease, task_group
                 "ORCHESTRATION_USER": self.config['provisioner_user'],
                 "ORCHESTRATION_HOST": self.config['provisioner_hostname']
             }
+        elif self.release.platform == "aws":
```
Harshith-umesh, line 64: this will only add the var for aws , if you want to add it for aws,gcp and azure for chaos, you could have an if condition like `if self.release.variant == "chaos":`

dags/openshift_nightlies/manifest.yaml
whitleykeith: they don't need to do this. They just need to make a variant in the cloud platform like acs did

whitleykeith (changes requested):
IMHO this is the wrong approach to this feature. Scripts like run_benchmark.sh are designed to be as simple as possible to avoid the "complex bash script" problems we've faced with other tools like e2e. We want to push as much of the orchestration logic (i.e. injecting aws creds, connecting to things,etc.) into the core Airflow code as it's (1) testable via our test suite and CI and (2) extensible and reusable. I would highly recommend investigating the other operators airflow provides (like the [KubernetesPodOperator](https://airflow.apache.org/docs/apache-airflow-providers-cncf-kubernetes/stable/operators.html)) and definitely look at making your own task class in this repo to actually handle the orchestration of whatever params you care about.

Some other nits I had:

1. Why do you need to run these commands on a jumphost? The pods that the benchmarks run in are in an internal cluster (sailplane).
2. Why do you need new install configs? You're adding a benchmark. I get making a new DAG but why not just reuse one of the sdn/ovn configs?

dags/openshift_nightlies/config/install/azure/chaos.json
```diff
@@ -0,0 +1,32 @@
+{
```
whitleykeith, line 1: these don't even work. Either make real configs or reference existing ones in the manifest.

dags/openshift_nightlies/scripts/run_benchmark.sh
```diff
@@ -81,14 +81,48 @@ if [[ $PLATFORM == "baremetal" ]]; then
...
+        echo "cd /root/kraken-hub/;podman run --env-file=container.env  --env-file=aws.env --name=application-outages --net=host  -v /root/.kube/config:/root/.kube/config:Z -d quay.io/openshift-scale/${chaos_test[$command]}"|ssh -o 'StrictHostKeyChecking=no' -o 'UserKnownHostsFile=/dev/null' -i ${PRIVATE_KEY} root@$CERBERUS_HOST /bin/bash
+        exit_code=`echo "podman inspect ${command} --format json"|ssh -o 'StrictHostKeyChecking=no' -o 'UserKnownHostsFile=/dev/null' -i ${PRIVATE_KEY}  root@$CERBERUS_HOST /bin/bash | jq .[].State.ExitCode`
+        if [[ 1 -ne $exit_code ]]; then
+            echo "podman logs -f ${command}"|ssh -o 'StrictHostKeyChecking=no' -o 'UserKnownHostsFile=/dev/null' -i ${PRIVATE_KEY} root@$CERBERUS_HOST /bin/bash
+            exit_code=`echo "podman inspect ${command} --format json"|ssh -o 'StrictHostKeyChecking=no' -o 'UserKnownHostsFile=/dev/null' -i ${PRIVATE_KEY} root@$CERBERUS_HOST /bin/bash | jq .[].State.ExitCode`
+        fi
+        echo "podman rm ${command}"|ssh -o 'StrictHostKeyChecking=no' -o 'UserKnownHostsFile=/dev/null' -i ${PRIVATE_KEY} root@$CERBERUS_HOST /bin/bash
+        benchmark_rv=$exit_code
```
whitleykeith, line 116: this seems unecessarily complex when all you need to do is run a container. Leverage airflow to do this and not bash. We only use bash here because the upstream tool (e2e) is in bash

dags/openshift_nightlies/tasks/benchmarks/e2e.py
whitleykeith: better yet just make a new task class for chaos benchmarks since this class is called `E2EBenchmarks` because it's for benchmarks specifically from https://github.com/cloud-bulldozer/e2e-benchmarking

dags/openshift_nightlies/config/install/azure/chaos.json
yogananth-subramanian: Forgot to double check and replace with config for azure and gcp. My apologies.

dags/openshift_nightlies/config/install/gcp/chaos.json
```diff
@@ -0,0 +1,32 @@
+{
```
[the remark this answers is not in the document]
yogananth-subramanian: Forgot to double check and replace with config for azure and gcp. My apologies.

dags/openshift_nightlies/tasks/install/openshift.py
```diff
@@ -106,7 +106,7 @@ def _get_playbook_operations(self, operation):
         else:
             return {"openshift_cleanup": True, "openshift_debug_config": False,
                     "openshift_install": False, "openshift_post_config": False, "openshift_post_install": False,
-                    "rhacs_enable": False}
+                    "rhacs_enable": False, "cerberus_enable": False}
```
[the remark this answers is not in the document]
yogananth-subramanian: If cerberus is going to be started as part of chaos task, then I guess scale-ci-deploy does not need to manage its installation and cleanup.

yogananth-subramanian:
> IMHO this is the wrong approach to this feature. Scripts like run_benchmark.sh are designed to be as simple as possible to avoid the "complex bash script" problems we've faced with other tools like e2e. We want to push as much of the orchestration logic (i.e. injecting aws creds, connecting to things,etc.) into the core Airflow code as it's (1) testable via our test suite and CI and (2) extensible and reusable. I would highly recommend investigating the other operators airflow provides (like the [KubernetesPodOperator](https://airflow.apache.org/docs/apache-airflow-providers-cncf-kubernetes/stable/operators.html)) and definitely look at making your own task class in this repo to actually handle the orchestration of whatever params you care about.
> 
> Some other nits I had:
> 
> 1. Why do you need to run these commands on a jumphost? The pods that the benchmarks run in are in an internal cluster (sailplane).
> 2. Why do you need new install configs? You're adding a benchmark. I get making a new DAG but why not just reuse one of the sdn/ovn configs?

Since the chaos tests are run in a pod, I was not sure if the Pods that the benchmarks run in support’s nested pods. Hence had used a jump host to spawn pods related to chaos tests. If patch is re-implement as separate task class using KubernetesPodOperator, then we can eliminate the use of jump host. 

scale-ci-deploy needed additional variable to install cerberus, hence had created a custom config. But if we are going to handle cerberus install in chaos task, we can eliminate the need to create separate config and reuse the existing one. 

Thank you for reviewing the patch. It was very helpful in realigning the scope and implementation of the feature.

Harshith-umesh:
Closing since this approach is outdated.

PR #121, opened 2022-02-08
Update Managed Services Runs to not to be concurrents

morenod:


smalleni (commented):
@morenod Is there some "airflow thingy" that prevents concurrent runs of a dag as well. Since users could manually trigger the same dag concurrently.

morenod:
> @morenod Is there some "airflow thingy" that prevents concurrent runs of a dag as well. Since users could manually trigger the same dag concurrently.

Concurrent builds of the same DAG are disabled by default on airflow. if somebody trigger a DAG when there is another one running, it will be waiting until the running one ends

smalleni:
> > @morenod Is there some "airflow thingy" that prevents concurrent runs of a dag as well. Since users could manually trigger the same dag concurrently.
> 
> Concurrent builds of the same DAG are disabled by default on airflow. if somebody trigger a DAG when there is another one running, it will be waiting until the running one ends

Gotcha, but what is someone triggers rosa ovn and rosa sdn at the same time. We would still run into the `osdCcsAdmin` credentials issue right?

dry923 (approved):
LGTM!

dry923:
@smalleni Mind giving this a quick review/merge please. The hope is that with STS we can work around the simultaneous install issue but its not in code yet.

whitleykeith (approved):
lgtm

smalleni (approved):

smalleni:
> @smalleni Mind giving this a quick review/merge please. The hope is that with STS we can work around the simultaneous install issue but its not in code yet.

I can merge if I'm given those permissions :)

PR #122, opened 2022-02-09
Add python 3.9, required by gcloud

morenod:
Exporting the var in the bashrc file is not enough, probably because airflow is not running from any shell. adding to the install script to force it usage


rsevilla87 (approved):
LGTM!

PR #125, opened 2022-02-10
Mark a dag as failed if any upstream tasks fail

Harshith-umesh:
### Description
Currently a dag is marked fail/success depending on the execution of the last cleanup task only. If any task fails upstream and if the cleanup task succeeds the dag is marked success since airflow only takes the status of the last task into account in a dag.
This PR adds a new last task called final_status which looks at the status of completed tasks and fails if any upstream task has failed, it fails logging the name of the task which failed. it passes only if all upstream tasks have passed successfully.
Eg of dag runs which fail and pass: http://harshith-umesh-faildag-airflow.apps.sailplane.perf.lab.eng.rdu2.redhat.com/tree?dag_id=4.10-aws-ovn-data-plane

### DIFF
--- a/dags/openshift_nightlies/tasks/utils/final_dag_status.py
+++ b/dags/openshift_nightlies/tasks/utils/final_dag_status.py
@@ -8,9 +8,15 @@ from airflow.operators.python import PythonOperator
 
 
 def final_status(**kwargs):
+    failed_tasks=[]
     for task_instance in kwargs['dag_run'].get_task_instances():
-        if task_instance.current_state() != 'success' and task_instance.task_id != kwargs['task_instance'].task_id:
-            raise Exception("Task {} failed. Failing this DAG run".format(task_instance.task_id))
+        if "index" in task_instance.task_id:
+            continue
+        elif task_instance.current_state() != 'success' and task_instance.task_id != kwargs['task_instance'].task_id:
+            failed_tasks.append(task_instance.task_id)
+
+    if len(failed_tasks) > 0:
+        raise Exception("Tasks {} failed. Failing this DAG run".format(failed_tasks))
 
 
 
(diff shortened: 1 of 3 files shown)


dags/openshift_nightlies/tasks/utils/final_dag_status.py
```diff
@@ -0,0 +1,24 @@
...
+from openshift_nightlies.models.dag_config import DagConfig
+
+
+from airflow.operators.python import PythonOperator
+
+
+def final_status(**kwargs):
+    for task_instance in kwargs['dag_run'].get_task_instances():
+        if task_instance.current_state() != 'success' and task_instance.task_id != kwargs['task_instance'].task_id:
+            raise Exception("Task {} failed. Failing this DAG run".format(task_instance.task_id))
```
whitleykeith, line 13: This is a bit nit picky here but instead of iterating over and raising an exception on the first task we see that failed, can the logic just grab all the failed tasks and then raise an exception if there are any? That way the log can log all the failed tasks in case more than 1 fails. 

i.e. 

```python
failed_tasks = [task_instance.task_id for task_instance in kwargs['dag_run'].get_task_instances() if task_instance.current_state == 'failed'] 

if len(failed_tasks) > 0:
  raise Exception(f" Tasks {failed_tasks} failed") 

```

```

whitleykeith (changes requested):
small nit but other than that lgtm

Harshith-umesh:
> @Harshith-umesh Does this currently ignore the indexing and /or cleanup status? Not sure if we would "care" if cleanup failed in the context of a successful run

this can be done , but the only time indexing and cleanup tasks fail is if the install task failed. both those run oc commands. so ignoring them or not may not make a difference i think, thoughts @whitleykeith ?

whitleykeith:
Indexing more or less only fails when the install fails (or the cluster under test is bored) since indexers need to hit that cluster to grab metadata.

Cleanup is a bit different though and can sometimes fail for reasons we may care about. If cleanups fail the cluster could get orphaned which would actually break future installs (this specifically happens a lot in AWS)

whitleykeith (approved):
lgtm!

PR #127, opened 2022-02-10
Add active waiting for rosa clusters to be ready

morenod:
Execution example: http://morenod-rosa-active-waiting-airflow.apps.sailplane.perf.lab.eng.rdu2.redhat.com/log?dag_id=4.10-rosa-ovn-control-plane&task_id=install&execution_date=2022-02-10T16%3A01%3A40.811817%2B00%3A00

### DIFF
--- a/dags/openshift_nightlies/scripts/install/rosa.sh
+++ b/dags/openshift_nightlies/scripts/install/rosa.sh
@@ -24,6 +24,28 @@ _get_cluster_status(){
     echo $(rosa list clusters -o json | jq -r '.[] | select(.name == '\"$1\"') | .status.state')
 }
 
+_wait_for_cluster_ready(){
+    echo "INFO: Waiting about 3 hour for cluster on ready status"
+    # 180 iterations, sleeping 60 seconds, 3 hours of wait
+    for i in `seq 1 180` ; do
+        CLUSTER_STATUS=$(_get_cluster_status $1)
+        if [ ${CLUSTER_STATUS} == "ready" ] ; then
+            sleep 120 # Wait 120 seconds to allow OCM to really finish the cluster installation
+	    echo "INFO: Cluster on ready status, dumping installation logs..."
+            rosa logs install -c $1
+    	    rosa describe cluster -c $1
+            return 0
+        else
+            echo "INFO: ${i}/180. Cluster on ${CLUSTER_STATUS} status, waiting 60 seconds for next check"
+            sleep 60
+        fi
+    done
+    echo "ERROR: Cluster $1 not installed after 3 hours, dumping installation logs..."
+    rosa logs install -c $1
+    rosa describe cluster -c $1
+    exit 1
+}
+
 setup(){
     mkdir /home/airflow/workspace
     cd /home/airflow/workspace
@@ -55,7 +77,8 @@ install(){
     export NETWORK_TYPE=$(cat ${json_file} | jq -r .openshift_network_type)
     export ROSA_VERSION=$(rosa list versions -o json --channel-group=nightly | jq -r '.[] | select(.raw_id|startswith('\"${version}\"')) | .raw_id' | sort -rV | head -1)
     [ -z ${ROSA_VERSION} ] && echo "ERROR: Image not found for version (${version}) on ROSA Nightly channel group" && exit 1
-    rosa create cluster --cluster-name ${CLUSTER_NAME} --version "${ROSA_VERSION}-nightly" --channel-group=nightly --multi-az --compute-machine-type ${COMPUTE_WORKERS_TYPE} --compute-nodes ${COMPUTE_WORKERS_NUMBER} --network-type ${NETWORK_TYPE} --watch
+    rosa create cluster --cluster-name ${CLUSTER_NAME} --version "${ROSA_VERSION}-nightly" --channel-group=nightly --multi-az --compute-machine-type ${COMPUTE_WORKERS_TYPE} --compute-nodes ${COMPUTE_WORKERS_NUMBER} --network-type ${NETWORK_TYPE}
+    _wait_for_cluster_ready ${CLUSTER_NAME}
     postinstall
 }
 


dags/openshift_nightlies/scripts/install/rosa.sh
```diff
@@ -24,6 +24,26 @@ _get_cluster_status(){
     echo $(rosa list clusters -o json | jq -r '.[] | select(.name == '\"$1\"') | .status.state')
 }
 
+_wait_for_cluster_ready(){
+    echo "INFO: Waiting about 1 hour for cluster on ready status"
+    # 180 iterations, sleeping 60 seconds, 3 hours of wait
+    for i in `seq 1 180` ; do
+        CLUSTER_STATUS=$(_get_cluster_status $1)
+        if [ ${CLUSTER_STATUS} == "ready" ] ; then
+            sleep 60 # Wait 60 seconds to allow OCM to really finish the cluster installation
```
smalleni, line 33: I've also seen that its usually way more than a minute after rosa cli reports cluster as ready, that the cluster is actually ready. So this might need to be bumped.

morenod: Moved to 2 mintes. anyway, this is just an skell for the function. after cluster is "ready" we should add more checks to verify if cluster is really ready or not. checking just the status returned by rosa is only the beginning of the active waiting, experience will give us the information about what we should check after that

morenod:
> @morenod How was the current behavior working? seems like we went to pstinstall as soon as the rosa create cluster command was fired?

Current installation command includes the --wait parameter, so command only ends when installation is "completed", according to the command. On active waiting, I remove the --wait parameter from the installation command, so it ends just when the cluster starts to install (it waits by default when it is on pending status), and then it is on our side, using the wait function to consider that installation is finished or not

dags/openshift_nightlies/scripts/install/rosa.sh
```diff
@@ -24,6 +24,26 @@ _get_cluster_status(){
...
+	    echo "INFO: Cluster on ready status, dumping installation logs..."
+            rosa logs install -c $1
+            return 0
+        else
+            echo "INFO: ${i}/60. Cluster on ${CLUSTER_STATUS} status, waiting 60 seconds for next check"
+            sleep 60
+        fi
+    done
+    echo "ERROR: Cluster $1 not installed after 3 hours, dumping installation logs..."
```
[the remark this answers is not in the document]
smalleni: OK, was wondering if this could be shorter maybe ~2hrs but 3 is fine. 

Do we use a different job to scale up? Then we should be fine setting lower timeout for failure since we would always only deploy smaller clusters to begin with.

morenod:
> > @morenod How was the current behavior working? seems like we went to pstinstall as soon as the rosa create cluster command was fired?
> 
> Current installation command includes the --wait parameter, so command only ends when installation is "completed", according to the command. On active waiting, I remove the --wait parameter from the installation command, so it ends just when the cluster starts to install (it waits by default when it is on pending status), and then it is on our side, using the wait function to consider that installation is finished or not

No, there are no scaleup jobs, clusters are installed always with the required number of workers. Sizing of master and infra nodes are selected depending on the number of workers. If we install with 3 workers, and then we scale over 25, sizing of masters and infra nodes will be wrong, causing the control-plane test results inconsistents

dry923:
> > > > @morenod How was the current behavior working? seems like we went to pstinstall as soon as the rosa create cluster command was fired?
> > > 
> > > 
> > > Current installation command includes the --wait parameter, so command only ends when installation is "completed", according to the command. On active waiting, I remove the --wait parameter from the installation command, so it ends just when the cluster starts to install (it waits by default when it is on pending status), and then it is on our side, using the wait function to consider that installation is finished or not
> > 
> > 
> > No, there are no scaleup jobs, clusters are installed always with the required number of workers. Sizing of master and infra nodes are selected depending on the number of workers. If we install with 3 workers, and then we scale over 25, sizing of masters and infra nodes will be wrong, causing the control-plane test results inconsistents
> 
> That is right, thanks for reminding David! So is 3 hrs good enough even for our largest clusters?

@smalleni our largest cluster currently in CI, probably. The 180 sets we haven't tried yet so thats TBD.

smalleni (approved):

dags/openshift_nightlies/scripts/install/rosa.sh
```diff
@@ -24,6 +24,26 @@ _get_cluster_status(){
     echo $(rosa list clusters -o json | jq -r '.[] | select(.name == '\"$1\"') | .status.state')
 }
 
+_wait_for_cluster_ready(){
+    echo "INFO: Waiting about 1 hour for cluster on ready status"
+    # 180 iterations, sleeping 60 seconds, 3 hours of wait
+    for i in `seq 1 180` ; do
+        CLUSTER_STATUS=$(_get_cluster_status $1)
+        if [ ${CLUSTER_STATUS} == "ready" ] ; then
+            sleep 120 # Wait 60 seconds to allow OCM to really finish the cluster installation
```
dry923, line 33: Comment is different than the sleep command

morenod: Updated all comments to new sleep values and counters

dags/openshift_nightlies/scripts/install/rosa.sh
```diff
@@ -24,6 +24,26 @@ _get_cluster_status(){
...
 
+_wait_for_cluster_ready(){
+    echo "INFO: Waiting about 3 hour for cluster on ready status"
+    # 180 iterations, sleeping 60 seconds, 3 hours of wait
+    for i in `seq 1 180` ; do
+        CLUSTER_STATUS=$(_get_cluster_status $1)
+        if [ ${CLUSTER_STATUS} == "ready" ] ; then
+            sleep 120 # Wait 120 seconds to allow OCM to really finish the cluster installation
+	    echo "INFO: Cluster on ready status, dumping installation logs..."
+            rosa logs install -c $1
```
dry923, line 35: @morenod one other request. Can you run a rosa describe cluster as well after the install? Has some useful information in there that might help if we need to debug.

morenod: done

morenod: ok, now I think it is really done

dry923:
@morenod any idea why the unit tests are failing?

dry923 (approved):
LGTM!

PR #136, opened 2022-03-11
Don't override METRICS_PROFILE

rsevilla87:

### Description

With the latest changes in e2e-benchmarking (https://github.com/cloud-bulldozer/e2e-benchmarking/pull/336) we no longer need to override METRICS_PROFILE (the default values are good) , just keeping the acs overrides and update it's expression


mukrishn (approved):
LGTM

PR #143, opened 2022-03-16
Fix unified kube-burner env vars

rsevilla87:

### Description
Fix unified kube-burner env vars. run_benchmark.sh does not work with notations like `WORKLOAD=node-density ./run.sh",`


morenod (approved):
LGTM

PR #145, opened 2022-03-16
Updating PODS_PER_NODE to 249

rsevilla87:
There're some jobs that are periodically created in clusters (kube-apiserver cert rotation, marketplace pods, registry image pruner, etc).
These jobs could be created while the workload is running leaving the nodes w/o room to run all workloads pods


morenod (approved):
LGTM

dry923 (approved):
LGTM

PR #149, opened 2022-03-18
cleanup stale gcp resources

mukrishn:
### Description
`ocm delete cluster` is leaving some resources behind on GCP provisioned cluster occasionally and that would fail the future iteration with `ERROR: (gcloud.compute.firewall-rules.create)`, so adding another `preclean` function to delete all stale resources from previous run. 
Also changing the `firewall-rules` name based on the network(they are unique) would also avoid duplicate names.

### DIFF
--- a/dags/openshift_nightlies/scripts/install/ocm_gcp.sh
+++ b/dags/openshift_nightlies/scripts/install/ocm_gcp.sh
@@ -87,11 +87,11 @@ postinstall(){
     ocm patch /api/clusters_mgmt/v1/clusters/$(_get_cluster_id ${CLUSTER_NAME}) <<< ${EXPIRATION_STRING}
     # Add firewall rules
     NETWORK_NAME=$(gcloud compute networks list --format="value(name)" | grep ^${CLUSTER_NAME})
-    gcloud compute firewall-rules create ${CLUSTER_NAME}-icmp --network ${NETWORK_NAME} --priority 101 --description 'scale-ci allow icmp' --allow icmp
-    gcloud compute firewall-rules create ${CLUSTER_NAME}-ssh --network ${NETWORK_NAME} --direction INGRESS --priority 102 --description 'scale-ci allow ssh' --allow tcp:22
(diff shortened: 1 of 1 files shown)


morenod (approved):
Awsome work, it totally LGTM!

dags/openshift_nightlies/scripts/install/ocm_gcp.sh
```diff
@@ -87,18 +87,36 @@ postinstall(){
...
+    gcloud compute firewall-rules create ${NETWORK_NAME}-ssh --network ${NETWORK_NAME} --direction INGRESS --priority 102 --description 'scale-ci allow ssh' --allow tcp:22
+    gcloud compute firewall-rules create ${NETWORK_NAME}-pbench --network ${NETWORK_NAME} --direction INGRESS --priority 103 --description 'scale-ci allow pbench-agents' --allow tcp:2022
+    gcloud compute firewall-rules create ${NETWORK_NAME}-net --network ${NETWORK_NAME} --direction INGRESS --priority 104 --description 'scale-ci allow tcp,udp network tests' --rules tcp:20000-20109,udp:20000-20109 --action allow
+    gcloud compute firewall-rules create ${NETWORK_NAME}-hostnet --network ${NETWORK_NAME} --priority 105 --description 'scale-ci allow tcp,udp hostnetwork tests' --rules tcp:32768-60999,udp:32768-60999 --action allow
 }
 
 cleanup(){
     ocm delete cluster $(_get_cluster_id ${CLUSTER_NAME})
     ocm logout
 }
```
Harshith-umesh, line 101: not sure how gcp/ocm works here but is it not possible to force this functionality at cleanup rather than running this as preinstall step. maybe save a few $$ by cleaning up the unused resources.

mukrishn: yes you are right, this should ideally be done by `ocm delete` command but there are many stale resources which are not cleaned up properly especially firewall-rules and network due to an existing issue(in 4.9), I think we are tracking the original issue and this patch is an additional workaround to clean them if any.

Harshith-umesh: ohh okay gotcha!

Harshith-umesh:
LGTM

dry923 (approved):
LGTM!

PR #151, opened 2022-03-22
Custom trigger rule

rsevilla87:
### Description

Permit setting custom trigger rules from the benchmark config file so we can move forward to the next task when the current task fails:

According to airflow documentation we can set these trigger rules:

```
    all_success: (default) all parents have succeeded
    all_failed: all parents are in a failed or upstream_failed state
    all_done: all parents are done with their execution
    one_failed: fires as soon as at least one parent has failed, it does not wait for all parents to be done
    one_success: fires as soon as at least one parent succeeds, it does not wait for all parents to be done
    none_failed: all parents have not failed (failed or upstream_failed) i.e. all parents have succeeded or been skipped
    none_skipped: no parent is in a skipped state, i.e. all parents are in a success, failed, or upstream_failed state
    dummy: dependencies are just for show, trigger at will
```

Also modified the control plane perfscale pipelines so that node-density expects the installation to be successful before being triggered  (all_success) and the subsequent tasks wait for its parents to be finished (all_done)

The final_status dag will work the same way as it used, and will report the failing tasks only.

----

Example where node-density failed and the subsequent tasks were executed regardless of this failure and final_status is reporting a failure in the dag:

image

### DIFF
--- a/dags/openshift_nightlies/tasks/utils/final_dag_status.py
+++ b/dags/openshift_nightlies/tasks/utils/final_dag_status.py
@@ -26,5 +26,6 @@ def get_task(dag):
         provide_context=True,
         python_callable=final_status,
         trigger_rule='all_done',
+        retries=0,
         dag=dag,
     )
--- a/dags/openshift_nightlies/config/benchmarks/control-plane.json
+++ b/dags/openshift_nightlies/config/benchmarks/control-plane.json
@@ -27,6 +27,7 @@
         {
             "name": "node-density-heavy",
             "workload": "kube-burner",
+            "trigger_rule": "all_done",
             "command": "./run.sh",
             "env": {
                 "WORKLOAD": "node-density-heavy",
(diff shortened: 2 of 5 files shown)


Harshith-umesh:
This is good in theory but what i'm worried about is if any of the kube-burner workloads fail and don't cleanup properly, will that be a problem?

rsevilla87:
> This is good in theory but what i'm worried about is if any of the kube-burner workloads fail and don't cleanup properly, will that be a problem? also if node-density fails what's the point of running node-density heavy?

The kube-burner cleanup mechanism is executed regardless the test fails or passes. As demostrated in the picture, running the remaining tests would have sense as they passed and node-density didn't.

dags/openshift_nightlies/tasks/utils/final_dag_status.py
```diff
@@ -26,5 +26,6 @@ def get_task(dag):
         provide_context=True,
         python_callable=final_status,
         trigger_rule='all_done',
+        retries=0,
```
Harshith-umesh, line 29: Good catch here raul, always wanted to add this but kept slipping my mind.

Harshith-umesh (approved):
LGTM!

dry923 (approved):
LGTM!

morenod (approved):
LGTM

PR #163, opened 2022-04-07
Remove some tasks and version 4.9

rsevilla87:

### Description

Let's speed up our pipelines by removing some tasks:

- baseline-performance: It's not worth to having this task. It's adding an extra 30 minutes delay and the value it provides is not very clear
- load_cluster_preupgrade and upgrades: These tasks don't work as they are expected, since the upgrade task tries to update to the latest version available, and the pipeline deploys the latest version available, for that reason, no upgrade is actually performed as shown in the log below:
```
[2022-04-07, 08:11:57 UTC] {subprocess.py:89} INFO - 2022-04-07T08:11:57Z - INFO     - MainProcess - py_es_bulk: Using streaming bulk indexer
[2022-04-07, 08:11:57 UTC] {subprocess.py:89} INFO - 2022-04-07T08:11:57Z - INFO     - MainProcess - wrapper_factory: identified upgrade as the benchmark wrapper
[2022-04-07, 08:12:04 UTC] {subprocess.py:89} INFO - 2022-04-07T08:12:04Z - INFO     - MainProcess - trigger_upgrade: Upgrading cluster mycluster to version 4.11.0-0.nightly-2022-04-01-172551 with uuid f78d7dd4-upgrades-20220407
[2022-04-07, 08:12:04 UTC] {subprocess.py:89} INFO - 2022-04-07T08:12:04Z - INFO     - MainProcess - trigger_upgrade: Current cluster version is: 4.11.0-0.nightly-2022-04-01-172551
[2022-04-07, 08:12:04 UTC] {subprocess.py:89} INFO - 2022-04-07T08:12:04Z - INFO     - MainProcess - trigger_upgrade: oc adm upgrade --to-latest=true --allow-upgrade-with-warnings=true --force=true
```

We have to rethink how we do this

### DIFF
--- a/dags/openshift_nightlies/config/benchmarks/control-plane.json
+++ b/dags/openshift_nightlies/config/benchmarks/control-plane.json
@@ -1,11 +1,6 @@
 {
     "benchmarks": [
         {
-            "name": "baseline",
-            "workload": "baseline-performance",
-            "command": "./baseline_perf.sh"            
-        },      
-        {
             "name": "node-density",
             "workload": "kube-burner",
             "command": "./run.sh",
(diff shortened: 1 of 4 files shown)


morenod (approved):
LGTM

dry923 (commented):
LGTM! I agree with the upgrade testing. That should probably be a one-off job where we specify a version to/from or if we want to automate it it needs to be an n-(Xversions) type situation

dry923 (approved):

Harshith-umesh:
@rsevilla87 reg baseline_perf, it's there so that it can identify if there's any problem with the base cluster itself, it captures prom metrics of an idle cluster. once there was a problem with the kube-api-server and this caught that and was used to analyze it. maybe we could remove it from regular-control-plane but keep them in medium and large ? it could be useful in the future if the need arises is all. cc: @mohit-sheth

Harshith-umesh:
LGTM!

PR #164, opened 2022-04-12
insecure es curl 

mukrishn:
### Description
To adopt new https ES endpoint,
```
curl: (60) SSL certificate problem: self signed certificate in certificate chain
```


rsevilla87 (approved):
lgtm!

PR #167, opened 2022-04-14
build osd mgmt cluster for hypershift

mukrishn:
### Description
Option to build AWS cluster using OSD.
This method will be used for building hypershift management cluster and can also be used for standalone AWS provided cluster as well using `"cluster_install_method": "osd"` in config json.


dags/openshift_nightlies/config/install/hypershift/node-type.json
```diff
@@ -0,0 +1,16 @@
+{
+    "cluster_install_method": "osd",
```
smalleni, line 2: Why do we need this node-type.json?

dags/openshift_nightlies/config/install/hypershift/rosa-ovn.json
```diff
@@ -0,0 +1,19 @@
+{
```
smalleni, line 1: While I appreciate that you have accounted for both ROSA and OSD management clusters, it's just too much code and complexity for something which does not fit with the deployment model. We know management clusters will be OSD only, lets stick to that and avoid complexity. My 0.02.

dags/openshift_nightlies/config/install/hypershift/sdn.json
```diff
@@ -1,10 +1,9 @@
 {
-    "aws_profile": "",
-    "aws_access_key_id": "",
-    "aws_secret_access_key": "",
-    "rosa_environment": "staging",
-    "rosa_cli_version": "master",
+    "cluster_install_method": "osd",
+    "ocm_cli_version": "master",
     "ocm_environment": "stage",
+    "rosa_environment": "staging",
```
smalleni, line 5: Do we need rosa specific vars for the OSD scenario as well? I don't think we use these vars in OSD so we must be able to skip these as they are only consumed it the cluster_install_method is not osd right?

dags/openshift_nightlies/scripts/install/rosa.sh
```diff
@@ -13,15 +13,23 @@ do
 done
```
smalleni, line 13: Since you do both ROSA and --provider aws for OSD using the same script, I wonder if we should change this to maanged_aws or something to avoid confusion, since you are clearly doing more than ROSA within this script.

dags/openshift_nightlies/scripts/run_benchmark.sh
```diff
@@ -12,7 +12,7 @@ setup(){
     mkdir /home/airflow/workspace
     cd /home/airflow/workspace
     echo "Cloning ${E2E_BENCHMARKING_REPO} from branch ${E2E_BENCHMARKING_BRANCH}"
-    git clone -b ${E2E_BENCHMARKING_BRANCH} ${E2E_BENCHMARKING_REPO} --depth=1 --single-branch
+    git clone -b master https://github.com/cloud-bulldozer/e2e-benchmarking.git --depth=1 --single-branch
```
smalleni, line 15: Is this intentional?

smalleni (changes requested):

dags/openshift_nightlies/scripts/run_benchmark.sh
mukrishn: This is for testing, I had to switch fork but will revert this before merge.

dags/openshift_nightlies/config/install/hypershift/sdn.json
mukrishn: yeah not required, will remove it.

dags/openshift_nightlies/config/install/hypershift/node-type.json
mukrishn: good catch, typo. should be `none-type.json` for no compute hosted cluster

dags/openshift_nightlies/scripts/install/rosa.sh
mukrishn: correct, I feel too. Either keep both rosa and osd under one umbrella `managed_aws` or keep rosa and osd_aws as separate platform. I can re-work accordingly. 
@dry923 @morenod what are your thoughts ?

dry923: I think merging them is fine TBH. They are virtually identical and we really don't even test straight OSD on AWS because we already do ROSA. Assuming that, I'd probably say we stick with ROSA since that is what most of the metadata will be tag'd with since we pull it from the system under test. Open to other opinions though.

dags/openshift_nightlies/scripts/install/hypershift.sh
```diff
@@ -43,30 +51,45 @@ setup(){
...
-    export ROSA_CLI_VERSION=$(cat ${json_file} | jq -r .rosa_cli_version)
-    if [[ ${ROSA_CLI_VERSION} == "master" ]]; then
-        git clone https://github.com/openshift/rosa
-	pushd rosa
-	make
-	sudo mv rosa /usr/local/bin/
-	popd
+    if [[ $INSTALL_METHOD == "osd" ]]; then
+        export OCM_CLI_VERSION=$(cat ${json_file} | jq -r .ocm_cli_version)
+        if [[ ${OCM_CLI_VERSION} == "master" ]]; then
```
dry923, line 56: So what happens if someone doesn't use the master branch? It looks like it doesn't get cloned down at all or am I missing something? Same thing for the below ROSA_CLI_VERSION as well

dags/openshift_nightlies/scripts/install/hypershift.sh
```diff
@@ -102,9 +124,8 @@ create_empty_cluster(){
     echo "Create None type Hosted cluster.."    
     export NETWORK_TYPE=$(cat ${json_file} | jq -r .hosted_cluster_network_type)
     export REPLICA_TYPE=$(cat ${json_file} | jq -r .hosted_control_plane_availability)   
-    BASEDOMAIN=$(_get_base_domain $(_get_cluster_id ${MGMT_CLUSTER_NAME}))
     echo $PULL_SECRET > pull-secret
-    hypershift create cluster none --name $HOSTED_CLUSTER_NAME --node-pool-replicas=0 --base-domain $BASEDOMAIN --pull-secret pull-secret --control-plane-availability-policy $REPLICA_TYPE --network-type $NETWORK_TYPE
+    hypershift create cluster none --name $HOSTED_CLUSTER_NAME --node-pool-replicas=0 --base-domain $BASEDOMAIN --pull-secret pull-secret --control-plane-availability-policy $REPLICA_TYPE --network-type $NETWORK_TYPE --control-plane-operator-image=quay.io/hypershift/hypershift:latest
```
dry923, line 128: We should probably add an option to point to a control-plane-operator-image based on a variable as well.

dags/openshift_nightlies/scripts/install/rosa.sh
```diff
@@ -52,37 +68,64 @@ setup(){
...
+        export OCM_CLI_VERSION=$(cat ${json_file} | jq -r .ocm_cli_version)
+        if [[ ${OCM_CLI_VERSION} == "master" ]]; then
+            git clone https://github.com/openshift-online/ocm-cli
+            pushd ocm-cli
+            sudo PATH=$PATH:/usr/bin:/usr/local/go/bin make
+            sudo mv ocm /usr/local/bin/
+            popd
+        fi
+        echo "Create new OSD access key.."
+        export ADMIN_KEY=$(aws iam create-access-key --user-name OsdCcsAdmin)
```
dry923, line 88: Since we are validating ocm w/ an ocm whoami It might make sense for us to also do an equivalent aws whoami to ensure our creds are working. This can be done via
```
aws iam get-user | jq .User.UserName
```
ex:
# aws iam get-user | jq .User.UserName
"rzaleski"

If you don't filter to the user name you will expose the userid and arn which we may not want in the logs.

dags/openshift_nightlies/scripts/install/hypershift.sh
mukrishn: This is just an optional override, the task [image](https://github.com/cloud-bulldozer/airflow-kubernetes/blob/master/images/airflow-managed-services/Dockerfile#L16-L17) got a stable version of CLI already.

smalleni: Yep, I couldn't find a way to do `if` conditions in a Dockerfile, so we have the stable version pre-installed and override the binary.

dry923: I guess my question was more. What if I get a test branch from someone that I need to test. It would be convenient if maybe I can just pass in the full git url or something as an override. Thoughts?

dags/openshift_nightlies/config/install/hypershift/sdn.json
mukrishn: my bad, they are required to get the token for `ocm login` [here](https://github.com/cloud-bulldozer/airflow-kubernetes/pull/167/files#diff-2077ba877e816fdbbd1f6ebea9c103d42861a79b2549fb3e93e16ae708c93e01R63)

dags/openshift_nightlies/scripts/install/rosa.sh
mukrishn: I am okay to keep them under ROSA for now unless @morenod has other opinions and work on it later if this cause any confusion.

dags/openshift_nightlies/scripts/install/hypershift.sh
mukrishn: Have we ever used any custom branch or fork for CLIs before or how often we do such overrides ? just one less variable to worry :)

dry923: So we have used custom branches for rosa during some early testing. My only thought here is that given hypershift is new and shiny we may end up in a position where we suddenly need a flag or something that isn't implemented yet. It also sets us up for the future. That said, I'm ok with no worrying about that in this PR and instead doing a followup to allow it for both rosa/ocm so we're covered in the future.

mukrishn: Okay lets do that, its good to have that provisioned and also for `hypershift` CLI.
I will work on all 3 of them on a separate pull request and keep this PR simple for OSD implementation only.

dags/openshift_nightlies/config/install/hypershift/none-type.json
```diff
@@ -0,0 +1,17 @@
+{
+    "cluster_install_method": "osd",
+    "ocm_cli_version": "master",
+    "ocm_environment": "stage",
+    "rosa_environment": "staging",
+    "rosa_expiration_time": "2880 minutes",
+    "openshift_worker_count": 9,
+    "openshift_network_type": "OpenShiftSDN",
+    "openshift_worker_instance_type": "m5.2xlarge",
```
dry923, line 9: We should default this to the current GA plan which is m6i.4xlarge, however, since m6i instances are scarce go w/ m5.4xl's

dags/openshift_nightlies/config/install/hypershift/none-type.json
```diff
@@ -0,0 +1,17 @@
...
+    "machineset_metadata_label_prefix": "machine.openshift.io",
+    "openshift_workload_node_instance_type": "m5.2xlarge",
```
dry923, line 11: are we actually using the workload node for the management cluster?

dags/openshift_nightlies/config/install/hypershift/none-type.json
dry923: same comments for the hypershift/ovn.json file as well

dags/openshift_nightlies/config/install/hypershift/none-type.json
mukrishn: I will remove workload spec.

dags/openshift_nightlies/config/install/hypershift/none-type.json
mukrishn: okay, assuming m6i is available now in all region, but still there may be changes coming as I am not sure if `hosted_cluster_instance_type`(guest worker) should be 4xl or 2xl.

dags/openshift_nightlies/scripts/install/rosa.sh
```diff
@@ -67,43 +106,55 @@ _wait_for_cluster_ready(){
...
+            # When detected a status change, index timer and update start time for next status change
+            DURATION=$(($CURRENT_TIMER - $START_TIMER))
+            INDEXDATA+=("${PREVIOUS_STATUS}"-"${DURATION}")
+            START_TIMER=${CURRENT_TIMER}
             echo "INFO: Cluster status changed to ${CLUSTER_STATUS}"
-	    if [ ${CLUSTER_STATUS} == "error" ] ; then
-	        rosa logs install -c $1
-                rosa describe cluster -c $1
+	        if [ ${CLUSTER_STATUS} == "error" ] ; then
+                if [[ $INSTALL_METHOD == "osd" ]]; then
```
morenod, line 117: This is going to fail, if else fi is not built correctly

dags/openshift_nightlies/scripts/install/rosa.sh
```diff
@@ -174,13 +271,20 @@ index_metadata(){
         _download_kubeconfig $(_get_cluster_id ${CLUSTER_NAME}) ./kubeconfig
         export KUBECONFIG=./kubeconfig
     fi
+    if [[ $INSTALL_METHOD == "osd" ]]; then
+        export PLATFORM="AWS"
```
morenod, line 275: we need to consider another platform name, as AWS is selected on benchmarks and tests as classic OCP on AWS (non managed). It could be confused when grouping clusters for results

dags/openshift_nightlies/scripts/install/rosa.sh
morenod: After checking all the code, I dont see too much shared code. I see a lot of conditionals like if [[ $INSTALL_METHOD == "osd" ]]; then, including all the code related to OSD and after that, the else containing the existing code for rosa.

After checking that, Im 75% on creating different installation scripts.

Its ok for this PR, but we should think on refactor when we have some time

morenod (changes requested):
@mukrishn please check some comments about a conditional not closed and the platform name

dags/openshift_nightlies/scripts/install/rosa.sh
mukrishn: good catch, lets do `AWSMS`  AWS Managed-Services ?

dags/openshift_nightlies/scripts/install/rosa.sh
mukrishn: Yeah I am okay to create a new script as a different platform(`awsms`) itself, but currently there are real switch-over only at 5 places(install, post-install, cleanup and 2 get commands) in `rosa.sh` to run `ocm create` cli commands instead of `rosa create` cli and rests are just env sets and logs so i thought of sharing the same script.  
Other changes you see in the PR are adding new json file for OSD AWS and `hypershift.sh` changes for OSD.

dags/openshift_nightlies/scripts/install/rosa.sh
morenod: OK, now I see it

morenod (approved):
LGTM

PR #168, opened 2022-04-19
Capturing installation times on rosa installations

morenod:
- New active waiting for all nodes to be ready after the installation process.
- Uploading installation times to openshift-install-timings elasticsearch index
   - All intermediate status until cluster is installing. Waiting, Validating and Pending
   - Installation : Time until cluster is on ready status when asking rosa api
   - Cluster Ready time: Time until all nodes are on ready status
   - Cleanup: Time to execute rosa delete cluster

### DIFF
--- a/dags/openshift_nightlies/scripts/install/rosa.sh
+++ b/dags/openshift_nightlies/scripts/install/rosa.sh
@@ -2,6 +2,8 @@
 
 set -ex
 
+export INDEXDATA=()
+
 while getopts v:a:j:o: flag
 do
     case "${flag}" in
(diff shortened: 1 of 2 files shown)


morenod:
Execution example: http://morenod-rosa-timing-airflow.apps.sailplane.perf.lab.eng.rdu2.redhat.com/log?dag_id=4.10-rosa-sdn-control-plane&task_id=install&execution_date=2022-04-18T16%3A08%3A32.687472%2B00%3A00
Data uploaded to ES: https://perf-results-kibana.apps.observability.perfscale.devcluster.openshift.com/app/discover#/?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now%2Fw,to:now%2Fw))&_a=(columns:!(),filters:!(),index:'9213e890-a42b-11ec-8892-4bcee73a8de1',interval:auto,query:(language:kuery,query:'platform.keyword:%20%22ROSA%22'),sort:!(!(timestamp,desc)))

smalleni (commented):
Where are we capturing the “Validating” state? We need to track that for the pre-flight checks done on the OCM end going forward (quota checks etc). Right now the Validating state is present only for a few seconds but going forward. Once the pre-flight checks are implemented it will be longer. Let’s start tracking that if we aren’t already.

morenod:
ES uploaded documents: https://perf-results-kibana.apps.observability.perfscale.devcluster.openshift.com/app/discover#/?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now%2Fd,to:now%2Fd))&_a=(columns:!(),filters:!(),index:'9213e890-a42b-11ec-8892-4bcee73a8de1',interval:auto,query:(language:kuery,query:'platform.keyword:%20%22ROSA%22%20and%20rosa_cluster_name.keyword:%20%22perf-410-d562%22'),sort:!(!(timestamp,desc)))

Installation log: http://morenod-rosa-timing-airflow.apps.sailplane.perf.lab.eng.rdu2.redhat.com/log?dag_id=4.10-rosa-sdn-control-plane&task_id=install&execution_date=2022-04-21T12%3A02%3A02.681374%2B00%3A00

dags/openshift_nightlies/scripts/install/rosa.sh
```diff
@@ -102,14 +166,48 @@ postinstall(){
...
+"aws_authentication_method": "${AWS_AUTHENTICATION_METHOD}",
+"cluster_version": "$(oc version -o json 2>/dev/null | jq -r .openshiftVersion)",
+"master_count": "$(oc get node -l node-role.kubernetes.io/master= --no-headers 2>/dev/null | wc -l)",
+"worker_count": "$(oc get node -l node-role.kubernetes.io/worker=,node-role.kubernetes.io/workload!=,node-role.kubernetes.io/infra!= --no-headers --ignore-not-found 2>/dev/null | wc -l)",
+"infra_count": "$(oc get node -l node-role.kubernetes.io/infra= --no-headers --ignore-not-found 2>/dev/null | wc -l)",
+"workload_count": "$(oc get node -l node-role.kubernetes.io/workload= --no-headers --ignore-not-found 2>/dev/null | wc -l)",
+"total_node_count": "$(oc get nodes 2>/dev/null | wc -l)",
+"ocp_cluster_name": "$(oc get infrastructure.config.openshift.io cluster -o json 2>/dev/null | jq -r .status.infrastructureName)",
+"rosa_cluster_name": "${CLUSTER_NAME}",
+"action": "${ACTION}",
```
smalleni, line 190: perhaps name this "state" to be consistent with what the rosa cli reports

smalleni:
> ES uploaded documents: https://perf-results-kibana.apps.observability.perfscale.devcluster.openshift.com/app/discover#/?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now%2Fd,to:now%2Fd))&_a=(columns:!(),filters:!(),index:'9213e890-a42b-11ec-8892-4bcee73a8de1',interval:auto,query:(language:kuery,query:'platform.keyword:%20%22ROSA%22%20and%20rosa_cluster_name.keyword:%20%22perf-410-d562%22'),sort:!(!(timestamp,desc)))
> 
> Installation log: http://morenod-rosa-timing-airflow.apps.sailplane.perf.lab.eng.rdu2.redhat.com/log?dag_id=4.10-rosa-sdn-control-plane&task_id=install&execution_date=2022-04-21T12%3A02%3A02.681374%2B00%3A00

Checked the docs at the ES link, looks good. We need a dashboard in grafana/Kibana with X-Axis having the ocp version and Y-Axis having stacked column charts to track this over releases. Would be cool to have this data captured over time to see if we are doing better/regressing.

smalleni:
@morenod Since some fields like the networkType, cluster name, cluster version, master_count and worker_count are only available when the cluster has been installed and not when its validating or installing, we would be missing important metadata to do run to run comparisons. Anyway we can capture that? (I know the oc command doesn't provide it before the cluster has been installed, but wondering if we can set some of these statically based on what was passed on the rosa cli. Thoughts?

morenod:
I'm not a fan of statics assignments, that's why I have added the Rosa cluster name which will be available on all documents of the same installation but also UUID will be also the same. Maybe we could wait until we build some dashboard associated with this timers and, if we have any problem showing the info, we can change to static assignments based on installation parameters

smalleni:
> I'm not a fan of statics assignments, that's why I have added the Rosa cluster name which will be available on all documents of the same installation but also UUID will be also the same. Maybe we could wait until we build some dashboard associated with this timers and, if we have any problem showing the info, we can change to static assignments based on installation parameters

I don't disagree with you but we atleast need the OCP version to be able to aggregate multiple runs so we can compare across versions over the dozens of runs we have. I think it's safe to atleast assign version because we don't really expect a different version to be installed than what was passed on the cli :)

dags/openshift_nightlies/scripts/install/rosa.sh
morenod: Perf team is already using action since time ago, if we want to compare results, we need to use the same parameter

morenod:
> > I'm not a fan of statics assignments, that's why I have added the Rosa cluster name which will be available on all documents of the same installation but also UUID will be also the same. Maybe we could wait until we build some dashboard associated with this timers and, if we have any problem showing the info, we can change to static assignments based on installation parameters
> 
> I don't disagree with you but we atleast need the OCP version to be able to aggregate multiple runs so we can compare across versions over the dozens of runs we have. I think it's safe to atleast assign version because we don't really expect a different version to be installed than what was passed on the cli :)

OK, so I have recheck all the parameters we are adding on the doc.

I have statically set platform, networktype, aws authentication, version and rosa cluster name, because they are parameters which wont change during or after the installation, and we now their value before the installation starts

Master and infra node counts cannot be set, they are dependant on the AWS region.
We could set workers and workload nodes, but they dont give important information on the phases that they cannot be calculated
OCP_cluster_name is rosa_cluster_name+4chars hash directly generated on OCP, we cannot set until we get from the cluster

dags/openshift_nightlies/scripts/install/rosa.sh
```diff
@@ -70,27 +132,28 @@ setup(){
     rosa whoami
     rosa verify quota
     rosa verify permissions
+    export ROSA_VERSION=$(rosa list versions -o json --channel-group=nightly | jq -r '.[] | select(.raw_id|startswith('\"${version}\"')) | .raw_id' | sort -rV | head -1)
```
smalleni, line 135: Does it make sense to capture a MAJOR_VERSION (4.9, 4.10) etc if we ever want to look holistically at major versions instead of the tiny details like minor version. If we see a major version regression then we can dig deeper into which nightly caused this. Thoughts?

morenod: For sure. $version parameter is pointing to the major version, I can add it on the metadata function

dags/openshift_nightlies/scripts/install/rosa.sh
```diff
@@ -100,17 +163,55 @@ postinstall(){
...
+        export KUBECONFIG=./kubeconfig
+    fi
+    local METADATA=$(cat << EOF
+{
+"uuid" : "${UUID}",
+"platform": "ROSA",
+"network_type": "${NETWORK_TYPE}",
+"aws_authentication_method": "${AWS_AUTHENTICATION_METHOD}",
+"cluster_version": "${ROSA_VERSION}",
+"cluster_major_version" : "${version}",
```
morenod, line 183: @sai added

dags/openshift_nightlies/scripts/install/rosa.sh
```diff
@@ -24,23 +24,81 @@ _get_cluster_status(){
...
+    echo "INFO: Waiting about 180 iterations, counting only when cluster enters on installing status"
+    ITERATIONS=0
+    PREVIOUS_STATUS=""
     # 180 iterations, sleeping 60 seconds, 3 hours of wait
-    for i in `seq 1 180` ; do
+    # Only increasing iterations on installing status
+    while [ ${ITERATIONS} -le 180 ] ; do
         CLUSTER_STATUS=$(_get_cluster_status $1)
+	CURRENT_TIMER=$(date +%s)
+	if [ ${CLUSTER_STATUS} != ${PREVIOUS_STATUS} ] ; then
```
dry923, line 69: and && ${PREVIOUS_STATUS} != ""

dags/openshift_nightlies/scripts/install/rosa.sh
```diff
@@ -24,23 +24,81 @@ _get_cluster_status(){
...
-    echo "INFO: Waiting about 3 hour for cluster on ready status"
+    START_TIMER=$(date +%s)
+    echo "INFO: Installation starts at $(date -d @${START_TIMER})"
+    echo "INFO: Waiting about 180 iterations, counting only when cluster enters on installing status"
+    ITERATIONS=0
+    PREVIOUS_STATUS=""
     # 180 iterations, sleeping 60 seconds, 3 hours of wait
-    for i in `seq 1 180` ; do
+    # Only increasing iterations on installing status
+    while [ ${ITERATIONS} -le 180 ] ; do
```
dry923, line 66: we could cut this time in half since a OCP install (with day 2 ops excluded) should not exceed 90 minutes for sure

dags/openshift_nightlies/scripts/install/rosa.sh
```diff
@@ -24,23 +24,81 @@ _get_cluster_status(){
...
         CLUSTER_STATUS=$(_get_cluster_status $1)
+	CURRENT_TIMER=$(date +%s)
+	if [ ${CLUSTER_STATUS} != ${PREVIOUS_STATUS} ] ; then
+		# When detected a status change, index timer and update start time for next status change
+                index_metadata ${START_TIMER} ${CURRENT_TIMER} ${PREVIOUS_STATUS}
```
dry923, line 71: After index_metadata. if status is errror exit 1. then we can remove the elif error from below

morenod: modified

dags/openshift_nightlies/scripts/install/rosa.sh
morenod: moved to 90

dags/openshift_nightlies/scripts/install/rosa.sh
morenod: added

dags/openshift_nightlies/scripts/install/rosa.sh
morenod: moved to managedservices-install-timings index

dry923 (approved):
LGTM! Great work!

PR #171, opened 2022-04-21
aws firewall rule for hostedcluster

### LINKED ISSUE #160: AWS fw rules for hosted cluster

mukrishn:
### Description
Post install tasks for hypershift HostedCluster that applies firewall/security-group policies on AWS for data plane traffic.

### DIFF
--- a/dags/openshift_nightlies/scripts/install/hypershift.sh
+++ b/dags/openshift_nightlies/scripts/install/hypershift.sh
@@ -96,6 +96,7 @@ create_cluster(){
     echo "Wait till hosted cluster is ready.."
     kubectl wait --for=condition=available --timeout=3600s hostedcluster -n clusters $HOSTED_CLUSTER_NAME
     postinstall
+    update_fw
 }
 
 create_empty_cluster(){
@@ -133,7 +134,6 @@ postinstall(){
             node=$(oc get nodes | grep worker | grep -i ready | wc -l)
             if [[ $node == $COMPUTE_WORKERS_NUMBER ]]; then
                 echo "All nodes are ready in cluster - $HOSTED_CLUSTER_NAME ..."
-                exit 0
             else
                 echo "Available node(s) are $node, still waiting for remaining nodes"
                 sleep 300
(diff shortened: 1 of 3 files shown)


mukrishn:
opened tcp rules up to `30109` to match with https://github.com/cloud-bulldozer/scale-ci-deploy/pull/183/files

dags/openshift_nightlies/scripts/install/hypershift.sh
```diff
@@ -147,6 +147,22 @@ postinstall(){
     fi
 }
 
+update_fw(){
```
smalleni, line 150: Are we doing this for ROSA?

smalleni: I'm not doing this on rosa or self-managed aws clusters, fwiw

dry923: @smalleni these are needed for uperf testing.

dry923 (approved):
LGTM!

dags/openshift_nightlies/scripts/install/hypershift.sh
smalleni: @dry923 ive done uperf testing without adding any of these on self-managed OCP on AWS. Are these needed for managed deployments?

mukrishn: @smalleni @dry923 Yes and no, these were needed for uperf testing on any manged deployment, but recently we [changed](https://github.com/cloud-bulldozer/benchmark-operator/pull/721/files#diff-c27a1bc402eddb25da91cc4c4072ed6b6760c8a9a742650fe5cff2022fec137f) uperf server ports to listen on `30000`  and there were following update made on [other](https://github.com/cloud-bulldozer/scale-ci-deploy/pull/183/files) places but  here. 
Fortunately there are default rules added by the AWS and GCP to include any port between `30000-32767` so uperf would work without it, but still I am keeping all other rules in place to match `scale-ci-deploy` policies. I can simply rules if needed LMK.

PR #173, opened 2022-04-26
Add rosa profiles for installing using AWS ccs authentication

morenod:
As we are using STS as default AWS authentication method on ROSA installations, we need to add profiles using CCS for comparing installation timings

### DIFF
--- a/dags/openshift_nightlies/config/install/rosa/ccs-ovn.json
+++ b/dags/openshift_nightlies/config/install/rosa/ccs-ovn.json
@@ -1,0 +1,14 @@
+{
+    "aws_profile": "",
+    "aws_access_key_id": "",
+    "aws_secret_access_key": "",
+    "aws_authentication_method": "ccs",
+    "rosa_environment": "staging",
+    "rosa_cli_version": "master",
+    "ocm_environment": "stage",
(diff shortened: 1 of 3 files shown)


morenod:
HOLD until https://github.com/cloud-bulldozer/airflow-kubernetes/pull/168 will be merge

smalleni (approved):
Looks good, do we want to rename the STS installs configs also like `sts-ovn.json` and `sts-sdn.json`. Perhaps not required since STS will be default going forward.

morenod:
> Looks good, do we want to rename the STS installs configs also like `sts-ovn.json` and `sts-sdn.json`. Perhaps not required since STS will be default going forward.

Yep, as it is the default, there is no need to identify as STS. My idea is to remove ccs-* runs after comparing results and left only the sts ones, with the common name

smalleni:
> > Looks good, do we want to rename the STS installs configs also like `sts-ovn.json` and `sts-sdn.json`. Perhaps not required since STS will be default going forward.
> 
> Yep, as it is the default, there is no need to identify as STS. My idea is to remove ccs-* runs after comparing results and left only the sts ones, with the common name

SGTM

PR #176, opened 2022-05-03
Do not select latest nightly, left one to upgrade after install

morenod:
In order to execute upgrade tests, we need to have at least one version to upgrade to, so we will install using the latest-1 version available


smalleni (approved):

morenod:
Closing this, as there is no upgrade options on nightly channel, something similar will be required for stable and candidate channels

PR #177, opened 2022-05-03
Removing kubeconfig on indexing cleanup for rosa

morenod:
Cleanup indexing is executed when cluster is already deleted, so no kubeconfig is available to download


dry923 (approved):
LGTM!

PR #180, opened 2022-05-05
Update airflow container images to 2.3.0

morenod:
After updating airflow to 2.3.0 version, we need to upgrade container images to the same version


Harshith-umesh:
Lets hold on this for now, 2.3.0 seems to have some issues, checking what we can do

Harshith-umesh (approved):
LGTM

PR #186, opened 2022-05-10
Task variables

rsevilla87:

### Description

Allows airflow to fetch arbitrary variables from specific dag/task tuple. This will helpful in case we want to, for example, configure different benchmark behaviors such as different metric profiles, comparison cofigurations, etc. between managed services and general perfscale DAGs.

Also deprecating the old overriding mechanism which turned out to be quite useless.

### DIFF
--- a/dags/openshift_nightlies/util/var_loader.py
+++ b/dags/openshift_nightlies/util/var_loader.py
@@ -15,19 +15,15 @@ def get_git_user():
     git_user = git_path.split('/')[0]
     return git_user.lower()
 
-def get_secret(name, deserialize_json=False):
-    overrides = get_overrides()
-    if name in overrides:
-        return overrides[name]
-    return Variable.get(name, deserialize_json=deserialize_json)
 
-
-def get_overrides():
-    try:
-        return Variable.get("overrides", deserialize_json=True)
-    except KeyError: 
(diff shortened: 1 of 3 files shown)


dags/openshift_nightlies/util/var_loader.py
```diff
@@ -15,19 +15,15 @@ def get_git_user():
...
-        return {}
-    
+def get_secret(name, deserialize_json=False, required=True):
+    if required:
+        return Variable.get(name, deserialize_json=deserialize_json)
+    else:
+        try:
+            return Variable.get(name, deserialize_json=deserialize_json)
+        except KeyError:
+            return {}
```
Harshith-umesh, line 27: why use if else above^? won't just try except be enough?

rsevilla87: The task variables are not required by default, so when required is False and the variable is not found in vault we return an empty dir

rsevilla87: It works, but I wanted to add the required logic to this method, so if one of the variables required is not present, the method will hang and not return an empty dir (which can be confusing)

Harshith-umesh (approved):
LGTM

PR #187, opened 2022-05-10
Execute remaining tasks when one fails

rsevilla87:

### Description

When one of the benchmarks fails we should continue the execution of the remaining


Harshith-umesh:
a quick question @rsevilla87 , like kube-burner will resources created from a previous workload(eg client,sever pods) be deleted before running the next workload?

Harshith-umesh:
a quick question @rsevilla87 , like kube-burner will resources created from a previous workload(eg client,sever pods) be deleted before running the next workload?

PR #188, opened 2022-05-12
adding ad-hoc capabilities

Harshith-umesh:
### Description
Fixing ad hoc run configs, https://github.com/cloud-bulldozer/airflow-kubernetes#ad-hoc-performance-tests. Enables user to select a specific nightly to use for a dag.

### DIFF
--- a/dags/openshift_nightlies/config/install/aws/acs.json
+++ b/dags/openshift_nightlies/config/install/aws/acs.json
@@ -2,6 +2,8 @@
     "aws_profile": "",
     "aws_access_key_id": "",
     "aws_secret_access_key": "",
+    "openshift_client_location": "",
+    "openshift_install_binary_url": "",
     "openshift_master_count": 3,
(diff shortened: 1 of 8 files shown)


Harshith-umesh:
@mohit-sheth @rsevilla87 i've added it only for aws for now , can add it for azure,gcp and alibaba if needed.

morenod:
> @mohit-sheth @rsevilla87 i've added it only for aws for now , can add it for azure,gcp and alibaba if needed.

Is this valid for other platforms? If so, can you add it on Managed Services ones?

Harshith-umesh:
> > @mohit-sheth @rsevilla87 i've added it only for aws for now , can add it for azure,gcp and alibaba if needed.
> 
> Is this valid for other platforms? If so, can you add it on Managed Services ones?

the last time we checked if this can be done for managed services, we concluded that it's not possible, but i don't remember why, let me check

PR #192, opened 2022-05-23
ROSA. Selecting channel group and OCP version to install

morenod:
- Add options for running rosa installation on all rosa channels (stable, candidate & nightly)
- Add option for selecting fixed OCP version on ROSA, latest image or prelatest for allowing ugrades.
- Removed schedule for ROSA CCS DAGS

### DIFF
--- a/dags/openshift_nightlies/manifest.yaml
+++ b/dags/openshift_nightlies/manifest.yaml
@@ -127,12 +127,12 @@ platforms:
           install: rosa/ovn.json
           benchmarks: data-plane.json
       - name: iam-ovn-control-plane
-        schedule: "None"
+        schedule: 'None'
         config:
           install: rosa/iam-ovn.json
           benchmarks: control-plane.json
       - name: iam-sdn-control-plane
-        schedule: "None"
+        schedule: 'None'
         config:
           install: rosa/iam-sdn.json
           benchmarks: control-plane.json          
--- a/dags/openshift_nightlies/scripts/install/rosa.sh
+++ b/dags/openshift_nightlies/scripts/install/rosa.sh
@@ -170,6 +170,8 @@ setup(){
     export AWS_AUTHENTICATION_METHOD=$(cat ${json_file} | jq -r .aws_authentication_method)
     export ROSA_ENVIRONMENT=$(cat ${json_file} | jq -r .rosa_environment)
     export ROSA_TOKEN=$(cat ${json_file} | jq -r .rosa_token_${ROSA_ENVIRONMENT})
+    export MANAGED_OCP_VERSION=$(cat ${json_file} | jq -r .managed_ocp_version)
+    export MANAGED_CHANNEL_GROUP=$(cat ${json_file} | jq -r .managed_channel_group)
     export CLUSTER_NAME=$(cat ${json_file} | jq -r .openshift_cluster_name)
     export COMPUTE_WORKERS_NUMBER=$(cat ${json_file} | jq -r .openshift_worker_count)
     export NETWORK_TYPE=$(cat ${json_file} | jq -r .openshift_network_type)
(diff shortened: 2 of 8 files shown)


morenod:
Installation example: http://morenod-managed-services-preflight-airflow.apps.sailplane.perf.lab.eng.rdu2.redhat.com/log?dag_id=4.10-rosa-sts-ovn-preflight&task_id=install&execution_date=2022-05-23T09%3A00%3A55.204697%2B00%3A00

Harshith-umesh:
Yay, so this logic worked out! 💯

smalleni (approved):
Good change, I like the fact that we can now pick the version we want to deploy specifically across channels.

mukrishn (approved):

dags/openshift_nightlies/config/install/rosa/preflight.json
```diff
@@ -0,0 +1,16 @@
...
+    "aws_profile": "",
+    "aws_access_key_id": "",
+    "aws_secret_access_key": "",
+    "aws_authentication_method": "sts",
+    "rosa_environment": "staging",
+    "rosa_cli_version": "master",
+    "ocm_environment": "stage",
+    "rosa_channel_group": "stable",
+    "rosa_ocp_version": "4.10.12",
+    "openshift_worker_count": 27,
```
dry923, line 11: Why are we testing this w/ 27 nodes for the install? I would think the minimum 3 would do given we're really just looking at the preflight times.

morenod: Totally right

dags/openshift_nightlies/manifest.yaml
```diff
@@ -144,6 +146,11 @@ platforms:
         config:
           install: rosa/ovn-osd.json
           benchmarks: data-plane.json
+      - name: sts-ovn-preflight
```
dry923, line 149: IMHO this should not be on a schedule.

dags/openshift_nightlies/scripts/install/rosa.sh
```diff
@@ -208,24 +210,37 @@ setup(){
...
     fi
 }
 
 install(){
     export COMPUTE_WORKERS_TYPE=$(cat ${json_file} | jq -r .openshift_worker_instance_type)
     if [[ $INSTALL_METHOD == "osd" ]]; then
-        export OCM_VERSION=$(ocm list versions --channel-group nightly | grep ^${version} | sort -rV | head -1)
-        [ -z ${OCM_VERSION} ] && echo "ERROR: Image not found for version (${version}) on OCM Nightly channel group" && exit 1
-        ocm create cluster --ccs --provider aws --region ${AWS_REGION} --aws-account-id ${AWS_ACCOUNT_ID} --aws-access-key-id ${AWS_ACCESS_KEY_ID} --aws-secret-access-key ${AWS_SECRET_ACCESS_KEY} --channel-group nightly --version ${OCM_VERSION} --multi-az --compute-nodes ${COMPUTE_WORKERS_NUMBER} --compute-machine-type ${COMPUTE_WORKERS_TYPE} --network-type ${NETWORK_TYPE} ${CLUSTER_NAME}
+	if [ "${ROSA_OCP_VERSION}" == "latest" ] ; then
```
dry923, line 231: Why are we using a ROSA_OCP_VERSION here and not OCM_VERSION?

morenod: Parameter on config files is rosa_ocp_version, it haven't been duplicated for OCM, as it points to an OCP version.
If pointed to a fixed version, script will check if version exists using ocm list version or rosa list version depending on the $cluster_install_method var, if points to latest, it will get the latest image available on $rosa_channel_group channel.

dags/openshift_nightlies/manifest.yaml
smalleni: @morenod I understand the executing automatically part, but this DAG doesn't need to be on a schedule on the production instance, can you not have the same exact effect by having the DAG in your fork/playground? To me, any testing that is short-term and one-off should be on playground and the more longer-term, continuous testing should live as a DAG on schedule on the production instance. What are your thoughts?

morenod: ok, removed

dags/openshift_nightlies/scripts/install/rosa.sh
dry923: If we are using the variable generically then we should change the name of it from ROSA.... to something else. MANAGED_OCP_VERSION perhaps?

morenod: modified to MANAGED_OCP_VERSION. in the same way I have changed ROSA_CHANNEL_GROUP to MANAGED_CHANNEL_GROUP

morenod:
@dry923 can you please review last changes? I have removed the DAG for preflights as it makes no sense because Im executing them on my airflow instance.

So now this PR is about allowing selecting the ROSA chanel group and the OCP version to be installed

dry923 (approved):
LGTM!

PR #190, opened 2022-05-19
Shrink data-plane cluster size

rsevilla87:
Signed-off-by: Raul Sevilla <rsevilla@redhat.com>

### Description

Several infrastructure changes, atm only affecting self-managed clusters :

- Use different configuration files for data-plane dags
- Master nodes size reduced by half, we don't actually require big master nodes in these dags
- Prometheus disk reduced from 500 to 50. These clusters are small and don't require a disk of this size
- AlertManager disk reduced from 20G to 1G. These clusters are small and don't require a disk of this size
- AlibabaCloud worker nodes now use cloud_ssd for data-plane nodes instead of cloud_essd
- Drop master nodes root disk size from 220 to 50 (I have verified that usage is less than 20GiB)
- Drop worker nodes root disk size from 100 to 50 (I have verified that usage is less than 20GiB)
- Drop workload nodes root disk size from 500 to 50 (I have verified that usage is less than 20GiB)

----

### Important changes in the infrastructure/workload node types:

Infrastructure nodes shifting from 48x192 to 16x64
Workload node shifting from 16x64 to 32x128

With the previous sizing, when running ingress-performance benchmarks, there was a big CPU bottleneck in the client side of the communication that leads to get not enough stable results (lot of variance between samples & benchmarks).

I verified that results are more stable after executing a few tests with this new node sizing. i.e: (Comparing 2 runs in  

```
+-----------+--------+----------------------+-----------+--------------------------+--------+-----------+--------------------------+--------------------------+
| test_type | routes | conn_per_targetroute | keepalive |          metric          | result | deviation | a4190995-router-20220519 | 35143df1-router-20220519 |
+-----------+--------+----------------------+-----------+--------------------------+--------+-----------+--------------------------+--------------------------+
|   http    |  500   |          1           |     0     | avg(requests_per_second) |  Pass  |   1.99%   |         51308.5          |         52329.0          |
|   http    |  500   |          1           |     1     | avg(requests_per_second) |  Pass  |  -1.71%   |         12528.5          |         12314.5          |
|   http    |  500   |          1           |    50     | avg(requests_per_second) |  Pass  |  10.11%   |         49962.5          |         55013.5          |
|   http    |  500   |          10          |     0     | avg(requests_per_second) |  Pass  |  -20.22%  |         111253.5         |         88762.0          |
|   http    |  500   |          10          |     1     | avg(requests_per_second) |  Pass  |   9.66%   |         20113.0          |         22055.0          |
[...]

PR #195, opened 2022-05-25
Reduce number of clients per route

rsevilla87:

### Description

Depends on https://github.com/cloud-bulldozer/airflow-kubernetes/pull/190

Transitioning from the default values:

LARGE_SCALE_CLIENTS: 1 20 80 -> 1 20 40
LARGE_SCALE_CLIENTS_MIX: 1 10 20 -> 1 5 10

The default values are excessive for the infrastructure proposed in #190 (and the older one as well), leading to CPU contention in the client side.

Also removing deprecated variables such as *_TOLERANCE and COMPARE, and reducing QUIET_PERIOD to 30s


sjug:
What's the benefit of changing the runtime?

rsevilla87:
> What's the benefit of changing the runtime? Do you have the CPU contention data handy that you're referring to?

There's no clear benefit of reducing RUNTIME, I reduced it because we're taking 4 samples now and didn't want to extend this test too much. Results are similar with RUNTIME 30 and 60 though

You can see CPU contention graphs in this other PR -> https://github.com/cloud-bulldozer/airflow-kubernetes/pull/190#issuecomment-1133209835  (This one actually depends on that one as well)

PR #198, opened 2022-05-26
Disable schedule on selected dags

morenod:
With current configuration, there is no way to disable schedules for a dag.
If there is no schedule line on the manifest, it is applied the default one.

This change allows to keep using default schedule if it is not defined on the dag, or disable for the dag, if it is defined as "None"

There are some dags without schedule, @rsevilla87 @mohit-sheth please review if you want to apply the default one, or the intention was to disable it
- [ACS](https://github.com/cloud-bulldozer/airflow-kubernetes/blob/master/dags/openshift_nightlies/manifest.yaml#L79)
- [Jetsky](https://github.com/cloud-bulldozer/airflow-kubernetes/blob/master/dags/openshift_nightlies/manifest.yaml#L88)
- [Openstack](https://github.com/cloud-bulldozer/airflow-kubernetes/blob/master/dags/openshift_nightlies/manifest.yaml#L93)
- [Prebuilt](https://github.com/cloud-bulldozer/airflow-kubernetes/blob/master/dags/openshift_nightlies/manifest.yaml#L181)

### DIFF
--- a/dags/openshift_nightlies/manifest.yaml
+++ b/dags/openshift_nightlies/manifest.yaml
@@ -127,10 +127,12 @@ platforms:
           install: rosa/ovn.json
           benchmarks: data-plane.json
       - name: iam-ovn-control-plane
+        schedule: "None"
         config:
           install: rosa/iam-ovn.json
           benchmarks: control-plane.json
       - name: iam-sdn-control-plane
+        schedule: "None"
         config:
           install: rosa/iam-sdn.json
           benchmarks: control-plane.json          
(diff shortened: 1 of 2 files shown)


dags/openshift_nightlies/manifest.yaml
```diff
@@ -128,9 +128,11 @@ platforms:
           benchmarks: data-plane.json
       - name: iam-ovn-control-plane
         config:
+        schedule: "None"
```
Harshith-umesh, line 131: schedule needs to be above config?

morenod: Yeah, failed c&p

rsevilla87 (approved):
Lgtm!

dry923 (approved):
LGTM

PR #204, opened 2022-05-31
Deploy minimum cluster and then scale-up

rsevilla87:

### Description

Deploy a small 3 node cluster and then scale-up in different steps for medium, large and extra-large dags. That will help us to collect scale-up metrics regularly


mohit-sheth (approved):
lgtm

PR #207, opened 2022-06-02
Use git user to fetch variables

rsevilla87:

### Description

Also adding missing docs for cluster_version override in BM environments.

### DIFF
--- a/dags/openshift_nightlies/docs/secrets.md
+++ b/dags/openshift_nightlies/docs/secrets.md
@@ -16,16 +16,23 @@ For users leveraging the `playground` or `tenant` mode of installation benefit f
 
 ## Variables for specific benchmark tasks
 
-It's possible to set arbitrary variables for a specific dag/task tuple, by default Airflow will try to fetch a secret with the name `<GIT_USER>-<DAG_NAME>-<TASK_NAME>` from a connected vault instance. The variables set in this secret will take precedence to any other variable defined in the task.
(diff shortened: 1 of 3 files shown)


smalleni (commented):
@rsevilla87 Should we include the branch name as well so we can have multiple palyground isntances per user?

smalleni:
> > @rsevilla87 Should we include the branch name as well so we can have multiple palyground isntances per user?
> 
> lol, you don't think `<GIT_USER>-<DAG_NAME>-<TASK_NAME>` is long enough? :-D, just kidding, I think is a good idea, we can use the original syntax `-<DAG_NAME>-<TASK_NAME>` for the main airflow instance and `<GIT_USER><GIT_BRANCH>-<DAG_NAME>-<TASK_NAME>` for a user's playgrounds

Sounds good, and how is this user specific secret populated in the vault instance?

rsevilla87:
> 

Same, let's say that you have a playground based in your fork and your branch ovn-stuff. You'd have to create a secret in vault with the name:

smalleni-ovn-stuff- 

> > > @rsevilla87 Should we include the branch name as well so we can have multiple palyground isntances per user?
> > 
> > 
> > lol, you don't think `<GIT_USER>-<DAG_NAME>-<TASK_NAME>` is long enough? :-D, just kidding, I think is a good idea, we can use the original syntax `-<DAG_NAME>-<TASK_NAME>` for the main airflow instance and `<GIT_USER><GIT_BRANCH>-<DAG_NAME>-<TASK_NAME>` for a user's playgrounds
> 
> Sounds good, and how is this user specific secret populated in the vault instance?

I updated the docs, take a look at them

PR #208, opened 2022-06-07
ocm - ocp nightly

venkataanil:
All rosa config files updated to use ocp version from nightly
channel group. This PR updates ocm rosa config file accordingly


chentex (approved):
LGTM

PR #211, opened 2022-06-13
Trigger rule for large-scale DAGs

rsevilla87:
### Description

Let's set a custom trigger_rule for the second benchmark of the large-scale tests, so in case cluster-density fails, we can execute node-density afterwards


Harshith-umesh (approved):
LGTM

PR #213, opened 2022-06-14
hypershift input changes

mukrishn:
### Description
1. Changes to Hypershift MC and HC install size, 
2. Operator configuration to enable metrics-set
3. Updated workload to use new P75 and P90
4. Updated release names per playground, added branch in the cluster name


dry923 (approved):
removed LGTM given our slack convo w/ you splitting the dags

venkataanil (approved):
LGTM

dags/openshift_nightlies/config/install/hypershift/none-type.json
```diff
@@ -1,9 +1,10 @@
 {
     "cluster_install_method": "osd",
-    "ocm_cli_version": "master",
```
morenod, line 3: I use to point to "container" on rosa when I dont want to use master, so we dont have on the default behavior. 

https://github.com/cloud-bulldozer/airflow-kubernetes/blob/master/dags/openshift_nightlies/scripts/install/rosa.sh#L201

It is better than comparing to null, which could be confuse

morenod (changes requested):

dags/openshift_nightlies/config/install/hypershift/none-type.json
mukrishn: Sure I will set a default value instead of `null`.

morenod (approved):
LGTM

PR #214, opened 2022-06-15
Change to use fixed version on rosa cli as latest has a bug

morenod:
Latest version of rosa cli has a bug on creating the machinepools, making the DAG fails.

Version 1.2.2 (last published) do not have the error, so we have to move config file to not to compile latest rosa cli


dry923 (approved):
LGTM!

PR #215, opened 2022-06-17
updated fw ports for uperf

mukrishn:
### Description
Uperf port allocation logic has been updated recently to be dynamic based on the number of pairs [here](https://github.com/cloud-bulldozer/benchmark-operator/pull/755/files#diff-12d1d181bb9b7ad8b89e717593b31d6ea210f5f205deb34c0a4c2852a5ebe0a5), so opening the ports wider to support that in AWS security groups and GCP FW rules

Good catch @vishnuchalla


dry923 (approved):
LGTM!

PR #218, opened 2022-06-27
Allow passing of install task variables to downstream tasks of a dag

Harshith-umesh:
### Description
This fix makes it possible to pass variables generated in the install task namely the cluster name and UUID of the install task for rosa and rogcp dags. The passed variables can be accessed by any of the following tasks(benchmarks,cleanup etc) in the dag.  
A limitation of the bash operator is that only a single string can be passed from a task(script), so a concatenation of the above 2 vars, separated by '@' delimiter is passed, the receiving tasks have to separate the 2  from the pulled string. 
Dag run eg: [dag link](http://harshith-umesh-checkvar-airflow.apps.sailplane.perf.lab.eng.rdu2.redhat.com/dags/4.11-rosa-iam-sdn-control-plane/graph?root=&execution_date=2022-06-27T09%3A46%3A05.574981%2B00%3A00) 
The passed string is accessible to the benchmark scripts and the cleanup script. 
### Fixes
https://issues.redhat.com/browse/PERFSCALE-1520


morenod (approved):
LGTM

PR #219, opened 2022-06-28
hypershift: loading dependencies var

mukrishn:
### Description
Included all env variables for hypershift benchmark execution and DAG dependency configs.


dry923 (approved):
LGTM!

PR #220, opened 2022-06-29
Connect airflow to postgres database

Harshith-umesh:
### Description
To connect airflow ci to the postgres database, UUID's of successful runs are written into one of the database tables. 
### Fixes
This is required for the data standardization/data science/data visualization work.


rsevilla87 (changes requested):
I'm not sure if this is the way to go. Inserting arbitrary stuff an external application database is usually not a good idea due to different reasons. i.e future airflow migrations could wipe all the not airflow tables. Thoughts?

rsevilla87 (approved):
lgtm

PR #221, opened 2022-07-01
PERFSCALE-1816 - Enables AWS Graviton2 tests.

afcollins:
### Description

Provisions airflow DAGs with AWS Graviton2 instances.

See matching branch on scale-ci-deploy of the same name.


afcollins:
scale-ci-deploy: https://github.com/cloud-bulldozer/scale-ci-deploy/pull/197

afcollins:
@mohit-sheth @rsevilla87 please review

dags/openshift_nightlies/docs/manifest_and_releases.md
```diff
@@ -50,12 +49,19 @@ platforms:
         config:
           install: openstack/sdn.json
           benchmarks: openstack.json
-
-
 ```
 
 would create a total of 8 dags ( 6 cloud-based, 2 openstack) with the naming convention being `$version-$provider-$variant`. This is because there is no fundamental parameter differences required between versions/providers, and this behavior gives us confidence that we're testing the same configurations across different versions and/or platforms.
 
+The children of the "config" section tell where under config to look for the definitions.
```
rsevilla87, line 56: Great to see improvements in the documentation :)

dags/openshift_nightlies/util/manifest.py
```diff
@@ -15,23 +18,36 @@ def __init__(self, root_dag_dir):
...
+                        "openshift_install_binary_url": f"{latest_accepted_release_url}/openshift-install-linux-{self.AMD64}-{latest_accepted_release}.tar.gz"
+                    }
+                else:
+                    # All other binaries under the "plain" ci URLs
+                    url = f"{self.release_stream_base_url}/{stream}/latest"
+                    latest_accepted_release,latest_accepted_release_url = self.request_for_payload(url)
+                    self.latest_releases[stream] = {
+                        "openshift_client_location": f"{latest_accepted_release_url}/openshift-client-linux-{latest_accepted_release}.tar.gz",
+                        "openshift_install_binary_url": f"{latest_accepted_release_url}/openshift-install-linux-{latest_accepted_release}.tar.gz"
+                    }
```
Harshith-umesh, line 44: could you please clarify about the function of the above for loop?

dags/openshift_nightlies/util/manifest.py
```diff
@@ -15,23 +18,36 @@ def __init__(self, root_dag_dir):
...
+        print(release_streams)
         self.latest_releases = {}
         for stream in release_streams:
-            url = f"{self.release_stream_base_url}/{stream}/latest"
-            response  = requests.get(url)
-            if response.status_code != 200: 
-                raise Exception(f"Can't get latest release from OpenShift Release API, API Returned {response.status_code}, {url}")
+            for arch in [ self.AMD64 , self.ARM64 ]:
+                if arch == self.ARM64:
+                    base_url_arm = self.release_stream_base_url.replace("openshift-release",f"openshift-release-{self.ARM64}")
```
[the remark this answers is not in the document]
Harshith-umesh: no i got that part about the different binaries, i was wondering if we could just use an if condition rather than if inside a for, but i figured that won't work. got it, thanks.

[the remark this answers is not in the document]
afcollins: Oh, you're right. the loop is not required and doesn't add anything meaningful.  I removed it and tests still pass. Thanks for pointing that out.

dags/openshift_nightlies/util/manifest.py
```diff
@@ -15,23 +18,34 @@ def __init__(self, root_dag_dir):
         self.release_stream_base_url = var_loader.get_secret("release_stream_base_url")
         self.get_latest_releases()
 
-    def get_latest_releases(self): 
+    def get_latest_releases(self):
         release_streams = [ version['releaseStream'] for version in self.yaml['versions']]
+        print(release_streams)
```
rsevilla87, line 23: I think we can remove this print message, or, replace it by log.info

rsevilla87 (approved):
LGTM

dags/openshift_nightlies/util/manifest.py
afcollins: Good catch. Will delete this one since it only prints what is in manifest.yaml. The releases that are resolved are printed elsewhere at: https://github.com/cloud-bulldozer/airflow-kubernetes/blob/master/dags/openshift_nightlies/dag.py#L296

dags/openshift_nightlies/util/var_loader.py
```diff
@@ -49,7 +49,7 @@ def get_config_vars(release: OpenshiftRelease, task="install", config_dir=f"{con
 
 def get_default_task_vars(release: OpenshiftRelease, task="install", task_dir=f"{constants.root_dag_dir}/tasks"):
     if task == "install":
-        if release.platform == "aws" or release.platform == "azure" or release.platform == "gcp" or release.platform == "alibaba":
+        if release.platform == "aws" or release.platform == "aws-arm" or release.platform == "azure" or release.platform == "gcp" or release.platform == "alibaba":
```
mohit-sheth, line 52: if release.platform in ("aws!", "aws-arm", .....):
we can use this pattern as we are adding more options

dags/openshift_nightlies/docs/manifest_and_releases.md
mohit-sheth: +1

mohit-sheth (commented):
lgtm, nice work @afcollins

dags/openshift_nightlies/util/var_loader.py
afcollins: Good call. Updated the condition, added few more tests, and refactored existing tests to be clearer (I hope).

PR #223, opened 2022-07-05
update airflow version to 2.3.2

Harshith-umesh:
### Description
updating airflow version references to 2.3.2 from 2.3.1.  Also updating tenet template which isn't up to date. 
### Fixes
The newer version fixes the grid display bug which is affecting a few dags.


rsevilla87 (approved):
lgtm!

PR #224, opened 2022-07-06
Removed the version from ansible 

mkarg75:
The ansible version we used has vanished from the repo. I've taken the version part away in the Dockerfile, so that we run with the latest at the time the images is being built. I have no idea how to ensure that jetski et.al work with that version, so this is a shot in the dark

### Description
removed ansible version 

### Fixes
image wasn't building since the package was not available


Harshith-umesh:
LGTM

Harshith-umesh (approved):
LGTM

PR #225, opened 2022-07-06
Only login on ocp when platform is ARO

morenod:
When working with managed services on gcp, username created cannot label nodes, so we exclude login on the cluster on this platform and use only kubeconfig


dry923 (approved):
LGTM!

rsevilla87 (approved):
lgtm

PR #230, opened 2022-07-11
Operator image

dry923:
### Description
We now need to pass a hypershift operator image to the hypershift installer. Additionally, it has been recomended that we do not define the control plane operator image anymore. We will leave the old variable, control_plane_operator_image, in there however defaulting it to "" results in the now desired functionality while giving us the ability to adjust it later if needed


dags/openshift_nightlies/config/install/hypershift/none-type.json
```diff
@@ -9,7 +9,8 @@
     "openshift_network_type": "OpenShiftSDN",
     "openshift_worker_instance_type": "m5.4xlarge",
     "number_of_hosted_cluster": 10,
-    "control_plane_operator_image": "quay.io/hypershift/hypershift:latest",
+    "control_plane_operator_image": "",
```
mukrishn, line 12: its the default set already [here](https://github.com/cloud-bulldozer/airflow-kubernetes/blob/master/dags/openshift_nightlies/tasks/install/hypershift/defaults.json#L36), so lets remove this from install input json files.

mukrishn (commented):
Look goods, just a small suggestion to remove `control_plane_operator_image` from input json.

PR #233, opened 2022-07-12
parallel cleanups fo hostedcluster

### LINKED ISSUE #232: [RFE] Deletion of Hypershift Hosted Clusters should be parellel
Deletion of a Hypershift Hosted Cluster can take 10+ minutes. We are currently doing this serially. That means that when I am doing tests of greater than 6 HCs it can take over an hour for cleanup. We should investigate doing this in a parallel format instead to greatly trim down the cleanup time.

mukrishn:


dry923 (approved):
Tested locally and it worked! LGTM!

PR #238, opened 2022-08-04
Use m6i instances for managed services

smalleni:
m6i instances have better performance/price and they now
seem to be enabled in ROSA for the regions we care about.
Leaving HyperShift at m5 instances and we can revisit based
on direction from HyperShift team.

Signed-off-by: Sai Sindhur Malleni <smalleni@redhat.com>

### DIFF
--- a/dags/openshift_nightlies/config/install/rosa/iam-ovn.json
+++ b/dags/openshift_nightlies/config/install/rosa/iam-ovn.json
@@ -10,7 +10,7 @@
     "managed_ocp_version": "latest",
     "openshift_worker_count": 27,
     "openshift_network_type": "OVNKubernetes",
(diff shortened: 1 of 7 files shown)


morenod (approved):
LGTM but check failed for any reason

mukrishn:
@smalleni LGTM, I assume then we gonna stick to m5 or m5a for hypershift MC ?

smalleni:
> @smalleni LGTM, I assume then we gonna stick to m5 or m5a for hypershift MC ?

Correct, the costs between m6.i and m5 are same. So, it doesnt really provide cost benefits but better perf/price. Let's wait on moving HyperShift, I pinged Naveen about it.

PR #239, opened 2022-08-05
Updated cluster name 

mukrishn:
### Description
replaced unsupported characters from the cluster name

### DIFF
--- a/dags/openshift_nightlies/models/release.py
+++ b/dags/openshift_nightlies/models/release.py
@@ -34,7 +34,7 @@ class OpenshiftRelease:
 
     def _generate_cluster_name(self):
         git_user = self._get_git_user()
-        git_branch = environ['GIT_BRANCH']
+        git_branch = str(environ['GIT_BRANCH']).replace("_","-").lower()
         release_name = self.get_release_name(delimiter="-")
         if git_user == 'cloud-bulldozer':
             cluster_name = f"ci-{release_name}"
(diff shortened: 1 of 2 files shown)


smalleni (approved):
LGTM

dags/openshift_nightlies/models/release.py
```diff
@@ -34,7 +34,7 @@ def _get_git_user(self):
 
     def _generate_cluster_name(self):
         git_user = self._get_git_user()
-        git_branch = environ['GIT_BRANCH']
+        git_branch = str(environ['GIT_BRANCH']).replace("_","-")
```
afcollins, line 37: Cool. Do we also need to lower-case any upper-case?

dags/tests/openshift_nightlies/util/test_manifest.py
```diff
@@ -59,7 +59,7 @@ def test_manifest_arm(self,mocked_manifest):
 
     def test_cloudreleases_amd(self,mocked_releases):
         releases = mocked_releases
-        assert len( releases ) == 145
+        assert len( releases ) == 143
```
afcollins, line 62: honestly this is a flakey assertion. You can probably remove both instead of changing.

jtaleric (approved):
lgtm

dags/openshift_nightlies/models/release.py
jtaleric: 🤔

dcbw: Yes, I think so.

jtaleric:
:shipit:

PR #242, opened 2022-08-09
checking for dashboard and updating messaging if its not there

radez:
The dashboard is not required for an airflow-kubernetes deployment.
If it's not there it can be ignored. This patch just checks if it's not there and prints a prettier message if it's not.


rsevilla87 (approved):
lgtm!

PR #244, opened 2022-08-15
Use default CNI

jtaleric:
We should accept the default CNI for OpenShift for ACS.

Signed-off-by: Joe Talerico (rook) <joe.talerico@gmail.com>


mohit-sheth (approved):
lgtm
fyi @gavin-stackrox

jtaleric:
Gavin gave his blessing on slack... merging..

PR #247, opened 2022-08-18
Hypershift cluster grafana-agent configuration

mukrishn:
### Description

- Removed deprecated hypershift CLI arguments 
- A custom grafana-agent configuration for Hypershift cluster to filter out metrics before scraping to avoid high memory consumption and preemption 
- Adding `workload`(r5.4xl) machinepool on management cluster and wait until its ready
- Enabling user-workload-monitoring stack to scrape HCP metrics as default, to keep CMO prometheus stable on 100 HC
- Configmap to pin user-workload-monitoring prometheus instances on `workload` nodes 
- moving grafana-agent pod to `workload` node


scripts/playground/templates/airflow.yaml
```diff
@@ -50,6 +50,8 @@ spec:
             killed_task_cleanup_time: 604800
             dag_file_processor_timeout: 300
             dagbag_import_error_traceback_depth: 4
+            parallelism: 200
+            max_active_tasks_per_dag: 200
```
mukrishn, line 54: I have increased to enable the concurrency when we go 100 HC.
the default value of `parallelism` is 32 per scheduler and `max_active_tasks_per_dag` is 16

dags/openshift_nightlies/tasks/utils/platform_connector.py
```diff
@@ -23,7 +23,8 @@ def __init__(self, dag, config: DagConfig, release: OpenshiftRelease, task_group
         self.env = {
             "REL_PLATFORM": self.release.platform,
             "THANOS_RECEIVER_URL": var_loader.get_secret("thanos_receiver_url"),
-            "LOKI_RECEIVER_URL": var_loader.get_secret("loki_receiver_url")
+            "LOKI_RECEIVER_URL": var_loader.get_secret("loki_receiver_url"),
```
smalleni, line 26: Is Loki being used? Not related to this PR but just curious?

mukrishn: yes, they are getting installed as part of this connector.

scripts/playground/templates/airflow.yaml
dry923: @mukrishn just for clarity, though, we may still be limited by the number of pods that airflow can spawn at one time. Since, to my understanding, each task is a new pod here.

mukrishn: Yes, correct each task spawn a new pod. if the cluster reaches its limit the task will be `scheduled` but pod would be pending state till kubelet schedules it. We have 3 workers and at max we can do 3x220 pods approximately, we should also adjust our cron schedule accordingly to avoid over committing tasks.

mukrishn:
@dry923 added sleep tasks after every parallel task, this will wait till all hostedclusters are installed then kick off benchmark concurrently on all HCs at same time and similarly cleanup. This way easy for us to kill and resume tasks manually.
Screenshot from 2022-08-26 17-06-13

mukrishn:
Ready for review, tested this for 100 HC.  
Screenshot from 2022-08-31 21-37-56

mukrishn:
@dry923  @morenod  Also, noticed creating all 100HC(~6000 pods) at once was slowing down the pod creation and all of them stuck in to a scheduling race condition. So to test this, I've added a skew logic to create HC once every minute(in [my fork](https://github.com/mukrishn/airflow-kubernetes/commit/c5e3648c3718b15b23f9409dde09d1111dcb4026#diff-2077ba877e816fdbbd1f6ebea9c103d42861a79b2549fb3e93e16ae708c93e01R54-R55) only). We should think of a permanent solution to over come this.

dry923 (approved):
LGTM

PR #248, opened 2022-08-22
Perf CI - Populate `use` tag so we can account for cloud spend better

jtaleric:
### Description
This will populate a tag `use` with the github-id of the user deploying an Airflow instance. 

### Need-to-do
Once this is merged we will need all current dags to be updated.


mohit-sheth:
LGTM

dry923 (approved):
LGTM!

PR #249, opened 2022-08-25
Set OCM endpoint testing rate and duration

venkataanil:
Set OCM endpoint testing rate and duration to original values.

I have reduced it earlier for quick testing while adding OCM
airflow dag. Now resetting it to original values for real testing.


chentex (approved):
LGTM

dry923 (approved):
LGTM!

PR #251, opened 2022-08-30
Fix python dependency versions to match airflow-2.3.2

afcollins:
### Description
Python dependencies used in this dockerfile would pull latest versions, which results in an incompatible combination that does not run.

### DIFF
--- a/images/airflow-jetski/Dockerfile
+++ b/images/airflow-jetski/Dockerfile
@@ -7,8 +7,7 @@ RUN dnf -y install python3.8 gcc platform-python-devel epel-release --nodocs
 RUN dnf --enablerepo=epel -y install sudo sshpass openssh-clients git ipmitool ansible python3-dns python3-netaddr ipmitool --nodocs
 
 RUN pip3.8 install update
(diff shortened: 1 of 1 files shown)


afcollins:
I saw numerous errors that go away after building an image with these fixed versions.

You might see any of the following:
```
ImportError: cannot import name 'ParameterSource' from 'click.core' (/usr/local/lib/python3.8/site-packages/click/core.py)
pkg_resources.VersionConflict: (WTForms 3.0.1 (/usr/local/lib/python3.8/site-packages), Requirement.parse('WTForms<3.0.0'))
AttributeError: module 'wtforms.fields' has no attribute 'TextField'
ERROR: flask 1.1.4 has requirement click<8.0,>=5.1, but you'll have click 8.1.3 which is incompatible.
ERROR: flask-appbuilder 3.4.5 has requirement WTForms<3.0.0, but you'll have wtforms 3.0.1 which is incompatible.
ERROR: redshift-connector 2.0.908 has requirement requests<2.28.1,>=2.23.0, but you'll have requests 2.28.1 which is incompatible.
```

PR #259, opened 2022-09-22
Update ocm-api-load to use --log-file

dry923:
### Description

Utilize the log-file option instead of redirecting to a file. This allows errors to still bubble up to the surface but keeps noise low.

### DIFF
--- a/dags/nocp/scripts/run_ocm_benchmark.sh
+++ b/dags/nocp/scripts/run_ocm_benchmark.sh
@@ -59,7 +59,7 @@ run_ocm_benchmark(){
 
     start_time=\$(date +%s)
     echo "Running the ocm-load-test test"
-    build/ocm-load-test --config-file config.yaml &> ${UUID}_ocm-load-test
+    build/ocm-load-test --config-file config.yaml --log-file ${UUID}_ocm-load-test
     cp ${UUID}_ocm-load-test /tmp/${UUID}/results/${UUID}_ocm-load-test.json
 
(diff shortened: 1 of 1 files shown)


dags/nocp/scripts/run_ocm_benchmark.sh
```diff
@@ -59,7 +59,7 @@ run_ocm_benchmark(){
 
     start_time=\$(date +%s)
     echo "Running the ocm-load-test test"
-    build/ocm-load-test --config-file config.yaml &> ${UUID}_ocm-load-test
+    build/ocm-load-test --config-file config.yaml --log-file ${UUID}_ocm-load-test
     cp ${UUID}_ocm-load-test /tmp/${UUID}/results/${UUID}_ocm-load-test.json
```
dry923, line 63: @venkataanil Can you explain why we are copying the log file to a new location and setting it as a .json? I couldn't find any reference to it after the copy and it isn't necessarily in json format. Thanks!

venkataanil: our intention here is to move ${UUID}_ocm-load-test log file to snappy server. I think automation.py doesn't pick it up if it is not a json file. I exactly don't know where this check is (might be at https://github.com/cloud-bulldozer/ocm-api-load/blob/main/automation.py#L485 ), however unable to push earlier when json extension is not added.

venkataanil (approved):
LGTM

chentex (approved):
LGTM

PR #261, opened 2022-09-29
Add support for running ARO DAG(s

krishvoor:
### Description

This PR adds support to running the ARO
1) Defined ARO releases via manifest.yaml
2) Scripts to install/clean up ARO


krishvoor:
Hi @morenod :wave:  
Would request your review and feedback

dags/openshift_nightlies/dag.py
```diff
@@ -1,3 +1,4 @@
+from sqlite3 import connect
```
morenod, line 1: is this being used anywhere? I cant see any reference to it

krishvoor: Nope, will clean it up

krishvoor: Updated.

morenod (approved):
LGTM

venkataanil (approved):
LGTM

dags/openshift_nightlies/manifest.yaml
```diff
@@ -182,6 +182,40 @@ platforms:
...
 
+  aro:
+    versions: ["4.10", "4.11"]
+    variants:
+      - name: sdn-control-plane
+        schedule:  "00 13 * * 3"
+        config:
+          install: azure/sdn-cp.json
+          benchmarks: control-plane.json
+      - name: ovn-control-plane
```
dry923, line 193: @Harshith-umesh did you test an OVN install? I don't see anything in the install code that captures which SDN to use. Additionally, with ARO I think you still need to jump through some hoops for OVN unless I'm mistaken.

krishvoor: Good find, I can install this [wheel](https://aka.ms/az-aroext-latest.whl) and pass in `--sdn-type`

```
    --sdn-type --software-defined-network-type    : SDN type either "OpenShiftSDN" (default) or
                                                    "OVNKubernetes".  Allowed values: OVNKubernetes,
                                                    OpenShiftSDN.
```

krishvoor: More Info on installing the preview is [here](https://learn.microsoft.com/en-us/azure/openshift/howto-configure-ovn-kubernetes#install-and-use-the-preview-azure-command-line-interface-cli)

Before and after installing the extension changes in parameter is [here](https://gist.github.com/krishvoor/83646514cca5baf58d6ad17549a9b2f3)

PR #266, opened 2022-10-25
added hc nodepool multi az arg

mukrishn:
### Description
Hypershift hostedclusters nodepool assignments are single zone by default, now adding an option to select multi az option and passing it to CLI arguments. 

##### Hypershift CLI
```
--zones strings       The availability zones in which NodePools will be created
```


dry923 (approved):
LGTM!

Issue #253, opened 2022-09-06
OCM dag failure: environment_new.txt file missing


venkataanil:
environment.txt file is missing though it is copied (dag log http://airflow.apps.sailplane.perf.lab.eng.rdu2.redhat.com/log?dag_id=ocm&task_id=api-load&execution_date=2022-08-26T00%3A00%3A00%2B00%3A00 )

[2022-09-01, 20:00:25 EDT] {subprocess.py:92} INFO - cat: /tmp/environment_new.txt: No such file or directory
[2022-09-01, 20:00:25 EDT] {subprocess.py:92} INFO - cat /tmp/environment_new.txt
[2022-09-01, 20:00:25 EDT] {subprocess.py:92} INFO - cat: /tmp/environment_new.txt: No such file or directory
[2022-09-01, 20:00:25 EDT] {subprocess.py:92} INFO - Creating aws key with admin user for OCM testing

Looks like https://github.com/cloud-bulldozer/airflow-kubernetes/blob/master/dags/nocp/scripts/run_ocm_benchmark.sh#L27 is removing all the files in /tmp folder.

This has happend earlier when dag is run through auto schedule (http://airflow.apps.sailplane.perf.lab.eng.rdu2.redhat.com/log?dag_id=ocm&task_id=api-load&execution_date=2022-08-26T00%3A00%3A00%2B00%3A00 ).

However not happend when triggered manually (http://airflow.apps.sailplane.perf.lab.eng.rdu2.redhat.com/log?dag_id=ocm&task_id=api-load&execution_date=2022-08-26T05%3A45%3A17.720087%2B00%3A00 ).

venkataanil:
Again manually triggered, which is successful http://airflow.apps.sailplane.perf.lab.eng.rdu2.redhat.com/log?dag_id=ocm&task_id=api-load&execution_date=2022-09-07T07%3A32%3A44.329811%2B00%3A00

chentex:
No longer valid

PR #267, opened 2022-11-08
fix ocm ci schedule

venkataanil:
Dag run is successful when triggered manually and failing only when auto scheduled. I got succesful run in my playground with these changes, so pushing a PR to run on airflow playground.

More details about this issue -
https://github.com/cloud-bulldozer/airflow-kubernetes/issues/253


dry923 (approved):
LGTM!

PR #270, opened 2022-11-14
hyp hosted cluster with external-dns

mukrishn:
### Description
By default `HostedCluster`s control plane traffic will get routed via management cluster router pods and that gets over utilized when running more clusters, on hosting 80 HC the usage increased up to 12-14 times actual requested pod resource and that might affect management cluster router performance as well.
This will be reduced if we have dedicated router pods for each HC, by adding `external-dns-domain` option will provision HA router pods within HCP namespace per HC. Hypershift when GA will do this by default when creating cluster using `rosa` cli. 
Adding this in to our automation to have all HCP pods for our management testing, `hc_external_dns` can be set only if hc is >= 4.12.


mukrishn:
@dry923 PTAL

dry923 (approved):
LGTM!

PR #272, opened 2022-11-17
hypershift mgmt kubeconfig for router-perf

mukrishn:
### Description
e2e-benchmarking router-perf script requires Hypershift Management Cluster access to modify default ingress and cluster version operator components when needed, passing env var to it.
Ref - https://github.com/cloud-bulldozer/e2e-benchmarking/pull/495


dry923 (approved):
LGTM!

PR #276, opened 2022-11-29
OSP update the storage class name as per 4.12

masco:
### Description
The default storage class name is changed from standard to standard-csi from 4.12, updating the same in the var file


Harshith-umesh (approved):
LGTM

PR #278, opened 2022-12-05
ocm: Allow autoschedule only for cloud-bulldozer

venkataanil:
Currently auto schedule is triggered for developer playgrounds as well and conflicting with prod dag run. We should follow openshift_nightlies dag approach and enable auto schedule only for prod dags i.e cloud-bulldozer git user

### DIFF
--- a/dags/nocp/util/manifest.py
+++ b/dags/nocp/util/manifest.py
@@ -1,6 +1,7 @@
 import yaml
 import requests
 from common.models.dag_config import DagConfig
+from nocp.util import var_loader
 
 class Manifest():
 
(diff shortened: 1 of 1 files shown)


venkataanil:
Unit test failures are not related to this change. We are seeing the same tox errors from Oct 10 (even for merged patches) https://github.com/cloud-bulldozer/airflow-kubernetes/pull/259/checks

All PRs in review are also failing with the same error https://github.com/cloud-bulldozer/airflow-kubernetes/actions/runs/3534349130/jobs/5931112165

I have reported https://github.com/cloud-bulldozer/airflow-kubernetes/issues/279 for the same now.

dry923 (approved):
LGTM!

PR #281, opened 2022-12-12
Custom cmd

rsevilla87:
### Description

Add logic to trigger custom commands rather than running only e2e-benchmarking based workloads. This will provide us much flexibility and freedom to include new benchmarks. 
Also updating base image to include latest kube-burner binary and adding some temporally example dags for testing purposes.

### DIFF
--- a/dags/openshift_nightlies/config/benchmarks/control-plane-custom.json
+++ b/dags/openshift_nightlies/config/benchmarks/control-plane-custom.json
@@ -1,0 +1,37 @@
+{
+    "benchmarks": [
+        {
+            "name": "scale-27",
+            "workload": "scale-perf",
+            "command": "./run_scale_fromgit.sh",
+            "env": {
+                "SCALE": "27",
+                "METADATA_COLLECTION": "true",
+                "WORKLOAD_NODE_ROLE": "workload"
+            }
+        },
+        {
+            "name": "node-density",
+            "workload": "kube-burner",
+            "custom_cmd": "kube-burner ocp node-density --uuid=${UUID} --pods-per-node=245 --timeout=2h --pod-ready-threshold=5s --es-server=${ES_SERVER} --es-index=${ES_INDEX}"
+        },
+        {
+            "name": "node-density-cni",
+            "workload": "kube-burner",
+            "trigger_rule": "all_done",
+            "custom_cmd": "kube-burner ocp node-density-cni --uuid=${UUID} --pods-per-node=245 --timeout=3h --es-server=${ES_SERVER} --es-index=${ES_INDEX}"
+        },
+        {
+            "name": "node-density-heavy",
+            "workload": "kube-burner",
+            "trigger_rule": "all_done",
+            "custom_cmd": "kube-burner ocp node-density-heavy --uuid=${UUID} --pods-per-node=245 --timeout=3h --es-server=${ES_SERVER} --es-index=${ES_INDEX}"
+        },
+        {
+            "name": "cluster-density",
+            "workload": "kube-burner",
+            "trigger_rule": "all_done",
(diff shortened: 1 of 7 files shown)


dags/openshift_nightlies/tasks/utils/final_dag_status.py
```diff
@@ -23,7 +23,6 @@ def final_status(**kwargs):
 def get_task(dag):
     return PythonOperator(
         task_id='final_status',
-        provide_context=True,
```
rsevilla87, line 26: This argument is deprecated, removing it.

dags/openshift_nightlies/scripts/run_benchmark.sh
```diff
@@ -78,12 +81,16 @@ echo "############################################"
...
     run_baremetal_benchmark
-    echo $UUID
 else
     setup
-    cd /home/airflow/workspace/e2e-benchmarking/workloads/$workload
+    if [[ -n ${workload} ]]; then
+        echo "Cloning ${E2E_BENCHMARKING_REPO} from branch ${E2E_BENCHMARKING_BRANCH}"
+        git clone -q -b ${E2E_BENCHMARKING_BRANCH} ${E2E_BENCHMARKING_REPO} --depth=1 --single-branch
+        cd /home/airflow/workspace/e2e-benchmarking/workloads/$workload
+    fi
```
rsevilla87, line 90: Only clone e2e-benchmarking repository when a specific workload is passed.

dags/openshift_nightlies/config/benchmarks/control-plane-custom.json
```diff
@@ -0,0 +1,37 @@
...
+            "env": {
+                "SCALE": "27",
+                "METADATA_COLLECTION": "true",
+                "WORKLOAD_NODE_ROLE": "workload"
+            }
+        },
+        {
+            "name": "node-density",
+            "workload": "kube-burner",
+            "custom_cmd": "kube-burner ocp node-density --uuid=${UUID} --pods-per-node=245 --timeout=2h --pod-ready-threshold=5s --es-server=${ES_SERVER} --es-index=${ES_INDEX}"
```
rsevilla87, line 16: The variables UUID, ES_SERVER and ES_INDEX are already automatically injected by airflow

dags/openshift_nightlies/scripts/run_benchmark.sh
```diff
@@ -97,6 +104,5 @@ else
         export WORKLOAD=$AIRFLOW_CTX_TASK_ID-must-gather
         ./ocp_diagnosis.sh > /dev/null
     fi
-    echo $UUID
```
mukrishn, line 100: was this intentional ? I think we need this for [indexing](https://github.com/cloud-bulldozer/airflow-kubernetes/blob/master/dags/openshift_nightlies/tasks/index/status.py#L51) job, unless if its printed by e2e or kube-burner scripts.

rsevilla87: woops, good catch, I wasn't aware of this trick, reverting this change

dags/openshift_nightlies/tasks/utils/final_dag_status.py
Harshith-umesh: @rsevilla87 this is required as we need to pass airflow context variables to kwargs of final_status.

rsevilla87: > @rsevilla87 this is required as we need to pass airflow context variables to kwargs of final_status.

https://airflow.apache.org/docs/apache-airflow/2.3.1/release_notes.html?highlight=provide_context#airflow-operators-python-pythonoperator

Harshith-umesh: ah okay!

Harshith-umesh (approved):
LGTM!

PR #284, opened 2022-12-19
Add new OCP 4.12 version to the DAGs

krishvoor:
Signed-off-by: Krishna Harsha Voora <krvoora@redhat.com>


dags/openshift_nightlies/config/install/hypershift/sdn.json
```diff
@@ -7,12 +7,12 @@
     "managed_channel_group": "nightly",
     "managed_ocp_version": "latest",    
     "openshift_worker_count": 9,
-    "openshift_network_type": "OpenShiftSDN",
```
mukrishn, line 10: this install config is meant for sdn type only, so instead we remove this dag.

dags/openshift_nightlies/manifest.yaml
```diff
@@ -126,7 +126,7 @@ platforms:
 
 # Do not program concurrent builds of ROSA/ROGCP/ARO
   rosa:
-    versions: ["4.10", "4.11"]
```
mukrishn, line 129: Also remove `4.10`

krishvoor: Done

dags/openshift_nightlies/config/install/hypershift/sdn.json
krishvoor: Agreed, removed

dags/openshift_nightlies/manifest.yaml
```diff
@@ -1,8 +1,4 @@
 versions:
-- version: "4.10"
```
mukrishn, line 2: this should not be removed, instead previous to be `4.11`, stable to `4.12` and next to `4.13` but Managed Service cluster has nothing to do with these `versions` so I'd leave this to self-managed team handle this rollout.

dags/openshift_nightlies/manifest.yaml
```diff
@@ -180,7 +176,7 @@ platforms:
           benchmarks: upgrade.json
 
   rogcp:
-    versions: ["4.10", "4.11"]
+    versions: ["4.10", "4.11", "4.12"]
```
mukrishn, line 179: missed this one

dags/openshift_nightlies/manifest.yaml
mukrishn: @krishvoor lets keep this and remove 4.10 from others too

mukrishn (approved):

PR #285, opened 2022-12-20
hypershift: corrected prefix of hostedcluster

mukrishn:
### Description
The workload env var isn't matching the right hostedcluster name [here](https://github.com/cloud-bulldozer/airflow-kubernetes/blob/master/dags/openshift_nightlies/scripts/install/hypershift.sh#L50)


mukrishn:
@krishvoor can you test with this change once ?

PR #286, opened 2022-12-21
Fix CI

rsevilla87:
### Description

New pip versions do a lot of backtracking, that could lead to extremely long dependency resolution, specially when there's a very complex dependency such as the kubernetes SDK.
To avoid so let's use fixed versions of the dependencies as suggested at the official docs
https://pip.pypa.io/en/stable/topics/dependency-resolution/#reduce-the-number-of-versions-pip-is-trying-to-use


Harshith-umesh (approved):
LGTM

PR #288, opened 2023-01-03
Spin up the OCP Clusters via ocm post 

krishvoor:
This PR adds relevant commands support to customize
the infra, compute, master nodes via ocm post command

Signed-off-by: Krishna Harsha Voora <krvoora@redhat.com>


dags/openshift_nightlies/scripts/install/rosa.sh
```diff
@@ -246,7 +248,13 @@ install(){
             export OCM_VERSION=$(ocm list versions --channel-group ${MANAGED_CHANNEL_GROUP} | grep ^${MANAGED_OCP_VERSION})
 	fi
         [ -z ${OCM_VERSION} ] && echo "ERROR: Image not found for version (${version}) on OCM ${MANAGED_CHANNEL_GROUP} channel group" && exit 1
-        ocm create cluster --ccs --provider aws --region ${AWS_REGION} --aws-account-id ${AWS_ACCOUNT_ID} --aws-access-key-id ${AWS_ACCESS_KEY_ID} --aws-secret-access-key ${AWS_SECRET_ACCESS_KEY} --channel-group ${MANAGED_CHANNEL_GROUP} --version ${OCM_VERSION} --multi-az --compute-nodes ${COMPUTE_WORKERS_NUMBER} --compute-machine-type ${COMPUTE_WORKERS_TYPE} --network-type ${NETWORK_TYPE} ${CLUSTER_NAME}
```
mukrishn, line 249: Just my thought, keep the original ocm create commands and include a new conditional block for custom master types when specified might be a better approach instead of doing this way. We won't be doing custom master deploy much often.

dry923: +1 I'd like to leave the default there as well and use the other option as a flaggable choice.

krishvoor: ACK will add a conditional block.

dags/openshift_nightlies/scripts/install/rosa.sh
```diff
@@ -246,7 +248,17 @@ install(){
...
+        
+        if [[ ${MASTER_MACHINE_TYPE}  == "m5.2xlarge" ]]; then
```
mukrishn, line 252: I would prefer another boolean flag `CUSTOM_MACHINE_TYPE` and leave it to `false` as default for regular ocm creation, because Master machine type is calculated based on the cluster size. https://docs.openshift.com/rosa/rosa_planning/rosa-limits-scalability.html#node-sizing-during-installation_rosa-limits-scalability

dags/openshift_nightlies/scripts/install/rosa.sh
```diff
@@ -246,7 +249,17 @@ install(){
...
-        ocm create cluster --ccs --provider aws --region ${AWS_REGION} --aws-account-id ${AWS_ACCOUNT_ID} --aws-access-key-id ${AWS_ACCESS_KEY_ID} --aws-secret-access-key ${AWS_SECRET_ACCESS_KEY} --channel-group ${MANAGED_CHANNEL_GROUP} --version ${OCM_VERSION} --multi-az --compute-nodes ${COMPUTE_WORKERS_NUMBER} --compute-machine-type ${COMPUTE_WORKERS_TYPE} --network-type ${NETWORK_TYPE} ${CLUSTER_NAME}
+        
+        if [[ ${CUSTOM_MACHINE_TYPE}  == "false" ]]; then
+            ocm create cluster --ccs --provider aws --region ${AWS_REGION} --aws-account-id ${AWS_ACCOUNT_ID} --aws-access-key-id ${AWS_ACCESS_KEY_ID} --aws-secret-access-key ${AWS_SECRET_ACCESS_KEY} --channel-group ${MANAGED_CHANNEL_GROUP} --version ${OCM_VERSION} --multi-az --compute-nodes ${COMPUTE_WORKERS_NUMBER} --compute-machine-type ${COMPUTE_WORKERS_TYPE} --network-type ${NETWORK_TYPE} ${CLUSTER_NAME}
+        else
+            echo "Cluster creation started"
+            curk -k https://gist.githubusercontent.com/krishvoor/34760584641e6bb06b75bcbdd076c639/raw/59918ebeb1b2dfedae3b57498c50c9a771232bf0/payload.json > payload_test.json
```
mukrishn, line 257: can this be kept local ? under `utils/templates` ?

PR #294, opened 2023-01-30
Wait form workload nodes only when configured

rsevilla87:

### Description

Airflow crashes when there's no workload nodes expected since it always tries to wait for them to show up. A conditional is missing to fix it.


morenod (approved):
LGTM

PR #298, opened 2023-02-01
Removing bad reference for gsheet key

jtaleric:
- Removing curl to server which is gone
- Moving gsheet key to vault
- Pulling from vault and storing key in same `/tmp/key.json`

Signed-off-by: Joe Talerico <jtaleric@redhat.com>


mohit-sheth:
lgtm!

jtaleric:
```shell
(airflow)ls /tmp | grep key
key.json```

Tested locally, and see the key in `/tmp/` which is the default location for the scripts to grab it from.

rsevilla87 (approved):
lgtm!

PR #307, opened 2023-04-10
pulled platform connector script out of workflow

mukrishn:
### Description
Removed the platform connector task from all DAGs, since the remote destination(thanos) has been decommissioned already.


rsevilla87 (approved):
LGTM!

PR #311, opened 2023-04-27
updated workflow

mukrishn:
### Description
To trigger a rebuild of airflow-images every night, to keep the CLIs up to date.

In ROSA build, we often get failures due to CLIs since they do frequent changes in the code, its better to keep the CLI up to date.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,10 +1,12 @@
 name: Release Airflow Images
 on:
+  schedule:  # triggers every midnight
+    - cron: '0 0 * * *' 
   push:
     branches:
       - master
     tags:
-      - "*" # triggers only if push new tag version
+      - "*" # triggers on a push event
 
 jobs:
   containers:


.github/workflows/release.yml
```diff
@@ -1,10 +1,12 @@
 name: Release Airflow Images
 on:
+  schedule:  # triggers every midnight
+    - cron: '0 0 * * *' 
   push:
     branches:
       - master
     tags:
-      - "*" # triggers only if push new tag version
+      - "*" # triggers on a push event
```
smalleni, line 9: Is the push section still needed when on schedule?

smalleni: @mukrishn

mukrishn: yes if any one makes a blocker fix and want to be reflected immediately, like the recent kube-burner version bump made many CI runs to fail.

rsevilla87 (approved):
lgtm!

PR #315, opened 2023-05-04
Trigger must-gather as soon as at least one parent has failed

rsevilla87:
### Description

The default value of the trigger_rule is `all_success`. [Docs](https://airflow.apache.org/docs/apache-airflow/2.3.2/concepts/dags.html#trigger-rules)

When one of the benchmarks fails we should must-gather the cluster from here:

This is what is happening now in tasks where MUST_GATHER_EACH_TASK is not `true`:
image
On the other hand, in case we merge this PR, do we need to must-gather from the run_benchmark.sh script?. I think they are redundant and confusing:

https://github.com/cloud-bulldozer/airflow-kubernetes/blob/ab1bbcd6a524dfc87c83ac0e6220f22b55d9a44e/dags/openshift_nightlies/scripts/run_benchmark.sh#L99-L109


jtaleric (approved):
lgtm

PR #317, opened 2023-05-05
Migrage cluster-density to ocp wrapper

rsevilla87:
Also removing the `--churn=true --churn-delay=2m --churn-duration=1h` flags, since they are defaulted,

And adding the `--user-metadata` flag to the commands in order to index the field `cpt: true` along with the workload metrics. cc: @jtaleric


jtaleric (approved):
lgtm!

PR #318, opened 2023-05-11
Update cleanup trigger rule to all_done

rsevilla87:
### Description

The PR https://github.com/cloud-bulldozer/airflow-kubernetes/pull/315 has a bug and leads to the deployed clusters not being pruned when all the upstream tasks from the DAG succeeded

image

We should execute cleanup always

Apart from that, I realized that must-gather only runs only when the upstream task fails, (where upstream means the directly preceding task). For that reason I've modified the code to add a must-gather task to each benchmark, so it gets executed only when that benchmark fails as I show below:

image


morenod (approved):
LGTM

PR #320, opened 2023-05-18
Fix final_dag task and add failed tasks detail to the slack message

rsevilla87:
### Description

Example showing two DAG runs, first one with a failed task and a second one w/o any failure:

image

Some linting as well.


jtaleric (approved):
lgtm

PR #321, opened 2023-05-23
Fix DAG trigger_rules

rsevilla87:
### Description

When a benchmark task fails must-gather should be executed and the subsequent benchmkarks should be skipped as well. 

There's an issue in the current implementation:

image

When must-gather is skipped, all the downstream tasks are skipped too. This happens because the DAG execution pipeline is sequential and does not have any branch

image

In the new implementation it has been updated to:

image

All benchmark's tasks trigger_rules are all_succeed, meaning that when a benchmark fails must-gather will run. Summarizing:

image

### DIFF
--- a/dags/openshift_nightlies/config/benchmarks/upgrade.json
+++ b/dags/openshift_nightlies/config/benchmarks/upgrade.json
@@ -3,8 +3,7 @@
         {
             "name": "cluster-density",
             "workload": "kube-burner",
-            "trigger_rule": "all_done",
-            "custom_cmd": "kube-burner ocp cluster-density --iterations=500 --timeout=2h -churn=false --gc=false"
+            "custom_cmd": "kube-burner ocp cluster-density --iterations=500 --timeout=2h --churn=false --gc=false"
         },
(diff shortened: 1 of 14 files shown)


dags/openshift_nightlies/config/benchmarks/upgrade.json
```diff
@@ -3,8 +3,7 @@
         {
             "name": "cluster-density",
             "workload": "kube-burner",
-            "trigger_rule": "all_done",
-            "custom_cmd": "kube-burner ocp cluster-density --iterations=500 --timeout=2h -churn=false --gc=false"
+            "custom_cmd": "kube-burner ocp cluster-density --iterations=500 --timeout=2h --churn=false --gc=false"
```
jtaleric, line 6: nice catch.

jtaleric (approved):
lgtm -- has this been tested with a playground?

rsevilla87:
> lgtm -- has this been tested with a playground?

Just finished running a full DAG using these bits

image

PR #326, opened 2023-06-05
Adding ingress-perf workload to data-plane-v2

rsevilla87:
### Description

Cluster is directly deployed with 9 worker nodes, as this is the worker count required to run the ingress-perf workload.

Infra nodes switched from the original (and needlessly) huge m5.12xlarge instances to c5.4xlarge (16x32)

After running several  benchmarks with multiple configurations, results with this configuration (available at https://github.com/cloud-bulldozer/e2e-benchmarking/pull/572)  have been quite stable as shown in the tale below based on the baseline UUID, this should allow us to detect regressions more efficiently.

image

Tolerancy has been set to 20%, meaning that a regression > 20% in one of the benchmark scenarios will mark the task as failed.


jtaleric (approved):
lgtm

PR #328, opened 2023-06-15
Remove cluster-density from DAGs

rsevilla87:
Running cluster-density-v2 is enough.

### DIFF
--- a/dags/openshift_nightlies/config/benchmarks/large-control-plane-mgs.json
+++ b/dags/openshift_nightlies/config/benchmarks/large-control-plane-mgs.json
@@ -3,12 +3,7 @@
         {
             "name": "node-density",
             "workload": "kube-burner",
-            "custom_cmd": "kube-burner ocp node-density --uuid=${UUID} --pods-per-node=245 --timeout=2h --pod-ready-threshold=10s --es-server=${ES_SERVER} --es-index=ripsaw-kube-burner --user-metadata=metadata.yml"
-        },
-        {
-            "name": "cluster-density",
-            "workload": "kube-burner",
-            "custom_cmd": "kube-burner ocp cluster-density --uuid=${UUID} --iterations=4000 --timeout=6h --es-server=${ES_SERVER} --es-index=ripsaw-kube-burner --user-metadata=metadata.yml"
+            "custom_cmd": "kube-burner ocp node-density --uuid=${UUID} --pods-per-node=245 --timeout=3h --pod-ready-threshold=10s --es-server=${ES_SERVER} --es-index=ripsaw-kube-burner --user-metadata=metadata.yml"
         },
         {
             "name": "cluster-density-v2",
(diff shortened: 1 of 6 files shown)


jtaleric:
Couple of questions - 
- Why 180 nodes?
- Why 3000 iterations? 

On the iterations, I am curious if this is just a random number we have always gone with? Nothing necessarily wrong with it, but if we are changing the number of iterations, how about we make it a calculation based off the number of nodes? This way we can have direct scaling numbers -- 24 nodes has 360 iterations, and 180 has 2700 -- if we go with 15 iterations. I know QE is using  `9*num_nodes` in Prow https://github.com/openshift/release/blob/master/ci-operator/step-registry/openshift-qe/cluster-density-v2/openshift-qe-cluster-density-v2-commands.sh#L23

rsevilla87:
> Couple of questions -
> 
>     * Why 180 nodes?
> 
>     * Why 3000 iterations?
> 
> 
> On the iterations, I am curious if this is just a random number we have always gone with? Nothing necessarily wrong with it, but if we are changing the number of iterations, how about we make it a calculation based off the number of nodes? This way we can have direct scaling numbers -- 24 nodes has 360 iterations, and 180 has 2700 -- if we go with 15 iterations. I know QE is using `9*num_nodes` in Prow https://github.com/openshift/release/blob/master/ci-operator/step-registry/openshift-qe/cluster-density-v2/openshift-qe-cluster-density-v2-commands.sh#L23

These are good questions.

- 180 nodes. I don't have a clear answer here, in the latest times we've been asked to run our scale tests using this number (and not 120), I thought that shifting the M cluster size was the way to go. thoughts @dry923 ?
---

- 3000 iterations?, as you suggested it's an arbitrary number and not linearly based on the cluster size. We can take advantage of this conversation to come up with a "formula".

Just to give some context cluster-density-v2 deploys 10 pods + 1 build per iteration. 
Using a ratio of 15 iterations per worker node would result in:

| Nodes | Iterations | ppn^ | bpn |
|--------|--------|--------|---|
| 24/S    | 360 | 150 | 15 |
| 180/M | 2700 | 150 | 15 |
| 252/XL | 3780 | 150 |15 |
| 501/XXL | **7515** | 150 |15 |

The number of iterations in the XXL is quite high, and it's very likely that the cluster (OVN?) won't support it .

jtaleric:
😄  

>The ratio you suggested 1:15, would result in:

15 was literally another random number 😭

dry923:
180 is the current max for managed rosa offerings hence why we have been pushing there lately (additionally its higher than what QE was looking at). 24->180 nodes is a large gap and jump, however if we assume that QE is covering this mid-range at 120 then we are good assuming we align the jobns.

For the cluster-density-ms job we worked out that 9*num nodes was a good representation of what customers using managed offerings are doing from previous telemetry data. The "normal" cluster-density jobs create far more objects than the ms variant so sticking to 9*nodes may still be more than enough.

jtaleric (approved):
lgtm

dry923 (approved):
LGTM!

